PR #11, opened 2023-06-03
NameRegistry, onboarding flow

dcposch:
@nalinbhardwaj


nalinbhardwaj (approved):
lgtm; might want to add an `owner` and override methods for them here so we can manual override if someone takes `vitalik` or something (at least while we're on testnet)

packages/daimo-api/src/model.ts
```diff
@@ -0,0 +1,16 @@
+import { Address } from "viem";
+import { z } from "zod";
```
nalinbhardwaj, line 2: cool library; im seeing it for the first time

dcposch:
ds

packages/daimo-api/gen/contract.ts
```diff
@@ -1,4 +1,4 @@
-// Generated by @wagmi/cli@1.1.0 on 6/3/2023 at 9:46:11 PM
+// Generated by @wagmi/cli@1.1.0 on 6/7/2023 at 3:04:12 AM
```
nalinbhardwaj, line 1: 👀that’s late haha🌃

PR #13, opened 2023-06-04
Ephemeral Notes contract

nalinbhardwaj:
On-chain, anyone who can show up with a signature of their address from the ephemeral (regular ECDSA) keypair can claim the note. Separately, the sender themselves can always claim the note (even without a signature), that way the sender can sweep old notes even if they lose the private keys themselves.

Off-chain, the expected flow is that the sender will send the private key (as part of the temp URL) to the receiver.

### DIFF
--- a/packages/contract/src/EphemeralNotes.sol
+++ b/packages/contract/src/EphemeralNotes.sol
@@ -10,6 +10,13 @@ struct Note {
     uint256 amount;
 }
 
+/**
+ * @title EphemeralNotes
+ * @notice Simple escrow notes, used for onboarding new accounts.
+ * Notes are created by a user and can be redeemed by anyone who shows
+ * a signature from the ephemeral key owner for their address. Additionally,
+ * the sender of the note can "revert" it without a signature.
+ */
 contract EphemeralNotes {
   mapping(address => Note) public notes;
   IERC20 public immutable token;
@@ -28,6 +35,8 @@ contract EphemeralNotes {
     uint256 _amount
   ) external {
     require(notes[_ephemeralOwner].ephemeralOwner == address(0), "EphemeralNotes: note already exists");
+    require(_ephemeralOwner != address(0));
+    require(_amount > 0);
 
     notes[_ephemeralOwner] = Note({
       ephemeralOwner: _ephemeralOwner,
@@ -39,6 +48,7 @@ contract EphemeralNotes {
     token.transferFrom(msg.sender, address(this), _amount);
   }
 
+  // Either the recipient (with a valid signature) or the sender can redeem the note
   function claimNote(
     address _ephemeralOwner,
     bytes memory _signature
--- a/packages/contract/src/EphemeralNotes.sol
+++ b/packages/contract/src/EphemeralNotes.sol
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.12;
+
+import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-contracts/contracts/utils/cryptography/ECDSA.sol";
+
+struct Note {
+    address ephemeralOwner;
+    address from;
+    uint256 amount;
+}
+
+contract EphemeralNotes {
+  mapping(address => Note) public notes;
+  IERC20 public immutable token;
+  
+  event NoteCreated(Note note);
+  event NoteRedeemed(Note note);
+
+  constructor(IERC20 _token) {
+    token = _token;
+  }
+
+
+  // Call token.approve(<address of this contract>, amount) on the token contract before this
+  function createNote(
+    address _ephemeralOwner,
+    uint256 _amount
+  ) external {
+    require(notes[_ephemeralOwner].ephemeralOwner == address(0), "EphemeralNotes: note already exists");
+
+    notes[_ephemeralOwner] = Note({
+      ephemeralOwner: _ephemeralOwner,
+      from: msg.sender,
+      amount: _amount
+    });
+
+    emit NoteCreated(notes[_ephemeralOwner]);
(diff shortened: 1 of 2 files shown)


packages/contract/src/EphemeralNotes.sol
```diff
@@ -0,0 +1,57 @@
...
+  
+  event NoteCreated(Note note);
+  event NoteRedeemed(Note note);
+
+  constructor(IERC20 _token) {
+    token = _token;
+  }
+
+
+  // Call token.approve(<address of this contract>, amount) on the token contract before this
```
dcposch, line 25: we can unlimited approve a few trusted contracts on account creation
- this one, for notes
- the paymaster, to pay fees in USDC

packages/contract/src/EphemeralNotes.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.12;
+
+import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-contracts/contracts/utils/cryptography/ECDSA.sol";
+
+struct Note {
+    address ephemeralOwner;
+    address from;
```
dcposch, line 9: storing `from` for convenience LGTM. it's redundant with logs, but L2 state is very cheap and the expensive part (calldata) is same either way.

packages/contract/src/EphemeralNotes.sol
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  function claimNote(
+    address _ephemeralOwner,
+    bytes memory _signature
+  ) external {
+    Note memory note = notes[_ephemeralOwner];
+    require(note.ephemeralOwner != address(0), "EphemeralNotes: note does not exist");
+
+    bytes32 message = ECDSA.toEthSignedMessageHash(keccak256(abi.encodePacked(msg.sender)));
```
dcposch, line 49: maybe dumb q: what is the advantage of signing keccak(sender) vs just signing sender

packages/contract/src/EphemeralNotes.sol
```diff
@@ -0,0 +1,57 @@
...
+    address signer = ECDSA.recover(message, _signature);
+    require(signer == _ephemeralOwner || msg.sender == note.from, "EphemeralNotes: invalid signature and not creator");
+
+    emit NoteRedeemed(note);
+    delete notes[_ephemeralOwner];
+    token.transfer(msg.sender, note.amount);
```
dcposch, line 55: 👍

packages/contract/src/EphemeralNotes.sol
```diff
@@ -0,0 +1,57 @@
...
+    token = _token;
+  }
+
+
+  // Call token.approve(<address of this contract>, amount) on the token contract before this
+  function createNote(
+    address _ephemeralOwner,
+    uint256 _amount
+  ) external {
+    require(notes[_ephemeralOwner].ephemeralOwner == address(0), "EphemeralNotes: note already exists");
```
dcposch, line 30: not strictly necessary, but it may also be worth checking

```
require(_ephemeralOwner != address(0));
require(_amount > 0);
```

packages/contract/test/EphemeralNotes.t.sol
```diff
@@ -0,0 +1,101 @@
...
+        assertEq(token.balanceOf(BOB), 500);
+        assertEq(token.balanceOf(address(notes)), 0);
+    }
+
+    function testFailingFlow() public {
+        // Alice approves token contract and creates a new note to ephemeralAddress
+        createAliceNote();
+
+        // Bob uses ephemeralAddress to sign wrong address and redeem the note
+        vm.startPrank(BOB, BOB);
```
dcposch, line 77: might throw in a `vm.expectRevert` above `claimNote`

packages/contract/test/EphemeralNotes.t.sol
```diff
@@ -0,0 +1,101 @@
...
+        assertEq(token.balanceOf(BOB), 0);
+        assertEq(token.balanceOf(address(notes)), 500);
+    }
+
+    function testRevertFlow() public {
+        // Alice approves token contract and creates a new note to ephemeralAddress
+        createAliceNote();
+
+        // Alice short circuits and reverts the note to herself
+        // She does not need a valid signature from ephemeralAddress to do this
```
dcposch, line 92: 👍

dcposch (approved):
LGreatTM

packages/contract/src/EphemeralNotes.sol
nalinbhardwaj: yeah not very necessary to do, mostly just makes client-side code for signatures a bit more straightforward since keccak256 are consistent bytes32, converting a `bytes memory` of address introduce slightly more complexity (padding, setting length in prefix correctly, string conversion etc.). can switch if you think just `sender` is cleaner.

packages/contract/src/EphemeralNotes.sol
nalinbhardwaj: Yeah, it's also stored so we are able to support the edge case of sender reverting unclaimed notes but the sender has lost the ephemeral private key themselves (can happen if for instance user deletes app and comes back years later or something)

packages/contract/src/EphemeralNotes.sol
nalinbhardwaj: added

packages/contract/test/EphemeralNotes.t.sol
nalinbhardwaj: added

packages/contract/src/EphemeralNotes.sol
dcposch: no keccak(sender) is great i was just curious

PR #17, opened 2023-06-10
Move contract types to their own package

nalinbhardwaj:
We need some of them for client-side app as well, for building userops and submitting transactions


nalinbhardwaj:
Also tried adding a test to make sure the generated file is never out of date with the latest run files, surprisingly it seems `wagmi generate` does not use stable sort for the file types (so there's always a diff)

dcposch:
I ran into the `wagmi generate` shuffle issue, too

dcposch:
### Filed an issue for `wagmi generate` stable output

https://github.com/wagmi-dev/wagmi/issues/2528

Quick response from the devs 🚀

packages/contract-types/src/index.ts
```diff
@@ -0,0 +1 @@
+export * as Contracts from "./generated";
```
dcposch, line 1: 🚀

packages/contract-types/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "module": "commonjs",
+    "esModuleInterop": true,
+    "target": "es2019",
+    "moduleResolution": "node",
+    "sourceMap": true,
+    "outDir": "dist",
+    "resolveJsonModule": true
+  },
+  "lib": ["es2019"]
+}
```
dcposch, line 12: you can use the node 20 preset:

```
{
  "extends": "@tsconfig/node20/tsconfig.json",
  "compilerOptions": {
    "strict": true,
    "resolveJsonModule": true,
    "sourceMap": true,
    "outDir": "dist"
  }
}
```

either way, most important is to add `strict: true`

packages/daimo-api/package.json
```diff
@@ -18,7 +17,7 @@
     "@tsconfig/node20": "^1.0.1",
     "@types/tape": "^5.6.0",
     "@wagmi/chains": "^1.1.0",
-    "@wagmi/cli": "^1.1.0",
+    "daimo-contract-types": "file:../contract-types",
```
dcposch, line 20: 🚀 

good call

dcposch (approved):
LGreatTM

packages/contract-types/tsconfig.json
dcposch: ^ i'll add it in a follup. merging now

nalinbhardwaj: fixed in https://github.com/daimo-eth/daimo/pull/24

PR #18, opened 2023-06-10
Upgrade account-abstraction to v0.6.0

nalinbhardwaj:
Based on #17, merge that first


packages/contract/src/Account.sol
```diff
@@ -22,12 +22,6 @@ import "./P256SHA256.sol";
 contract Account is BaseAccount, UUPSUpgradeable, Initializable {
     using ECDSA for bytes32;
 
-    //filler member, to push the nonce and owner to the same slot
-    // the "Initializeble" class takes 2 bytes in the first slot
-    bytes28 private _filler;
-
-    //explicit sizes of nonce, to fit a single storage cell with account keys
-    uint96 private _nonce;
```
dcposch, line 30: 👍

packages/contract/src/Account.sol
```diff
@@ -171,7 +153,7 @@ contract Account is BaseAccount, UUPSUpgradeable, Initializable {
     function withdrawDepositTo(
         address payable withdrawAddress,
         uint256 amount
-    ) public onlyOwner {
+    ) public onlySelf {
```
dcposch, line 156: I'm not sure if we need deposits

If we use a paymaster
- Non-upgradeable, vetted paymaster contract
- Unlimited approve the paymaster on account creation
- All ops subsequently paid by the paymaster

...do we still need this mechanism for depositing to the entrypoint?

(We don't have to address on this PR.)

dcposch (approved):
LGTM

packages/contract/src/Account.sol
nalinbhardwaj: One edge case that we should think about is if the underlying AA spec gets breaking upgrades and we need to upgrade in sync of those. For instance, the latest upgrade removed the nonce handling from the account itself to the entry point code, (I believe) breaking old deployed accounts. That's mostly why I think having a withdraw is helpful, we can "upgrade" accounts by making a new contract + withdrawing deposit to it if necessary.

nalinbhardwaj: It's low cost to keep here anyway for now so will punt this delta.

PR #24, opened 2023-06-11
contract-types: Use Node 20 tsconfig preset

nalinbhardwaj:
Review comment


packages/contract-types/dist/contract-types/src/index.d.ts
```diff
@@ -0,0 +1 @@
+export * as Contracts from "./generated";
```
nalinbhardwaj, line 1: @dcposch do you know if there's a way to not commit these dist files but build them for other package includes? do we just need to add a publish pipeline to npm or something for that?

I think some yarn/npm workspace magic should also be able to do this automagically, but I can't find on googling easily

packages/contract-types/package.json
```diff
@@ -2,17 +2,18 @@
   "name": "daimo-contract-types",
   "version": "1.0.0",
   "description": "Typescript types and ABIs for daimo contracts",
-  "main": "src/index.ts",
+  "types": "./dist/contract-types/src/index.d.ts",
+  "main": "./dist/contract-types/src/index.js",
```
dcposch, line 6: was something breaking linking to `src/index.ts`?

Building typescript is generally necessary when publishing a library to NPM, but if we're only using it locally from another typescript project, I'm not sure

nalinbhardwaj: yeah https://github.com/daimo-eth/daimo/pull/25#discussion_r1225900978 requires us to `npm pack` these which only picks from the built files.

dcposch:
combined into #26

PR #32, opened 2023-06-14
Implement Send

dcposch:
ds


apps/daimo-mobile/src/logic/chain.ts
```diff
@@ -115,6 +147,45 @@ export class ViemChain implements Chain {
...
+    const onLogs = function (logs: TransferLog[]) {
+      console.log(`[CHAIN] got ${logs.length} transfers for ${address}`);
+      for (const log of logs) {
+        if (!log.args.from || !log.args.to || !log.args.value) {
+          console.warn(`[CHAIN] skipping invalid Tranfer log ${log}`);
+          continue;
+        }
+
+        const wei = 10 ** coin.decimals;
+        const amount = (Number(log.args.value) / wei).toFixed(2);
```
nalinbhardwaj, line 161: nit I think [`formatUnits`](https://viem.sh/docs/utilities/formatUnits.html) can also do this

apps/daimo-mobile/src/logic/chain.ts
```diff
@@ -1,33 +1,45 @@
...
 export const chainConfig = {
   testnet: true,
   l1: goerli,
   l2: baseGoerli,
-  coinContract: "0x1B85deDe8178E18CdE599B4C9d913534553C3dBf" as Address,
 };
 
+export const coin = {
+  address: "0x1B85deDe8178E18CdE599B4C9d913534553C3dBf" as Address,
+  decimals: 6,
```
nalinbhardwaj, line 29: nit and dont need to do here but I was thinking we can make @daimo/contracts generate and export some of the metadata of the erc20 contract separately like `address`/`decimals`/`name`/`symbol` etc. so we don't have to hardcode or fetch in-app. Will also make simpler to swap it out in future for us (and others who fork)

dcposch:
### User op send issue

ds

all addresses (entry point, USDC, destination) look valid

apps/daimo-mobile/src/logic/chain.ts
dcposch: formatUnits doesn't have display decimals (so might output `"42.690418"` for 6-decimal USDC)

dcposch:
### Fixed; send works now

TLDR;
- **To a send a user op, we need gas.** Either prepaid (ETH deposited in the EntryPoint on behalf of this specific account) or pay as you go (with a paymaster). 
- **We'll switch to paymaster once ready.** Kristof shipping a paymaster on Base soon. Package `@pimlico/erc20-paymaster` looks promising, still in active development.
- **See `scratchpad` package.** This is an easy place to try transactions, crypto, contract interactions without running a full Expo app.

apps/daimo-mobile/src/action/actStatus.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** Action status */
+  status: ActStatus;
+  /** Empty when idle. Describes progress, success, or failure. */
+  message: string;
+  /** Should be called only when status is 'idle' */
+  exec: () => void;
+}
+
+/** Tracks progress of a user action. */
+export function useActStatus() {
```
dcposch, line 20: this is a nice reusable pattern

react hook which is meant to track a single user action, such as
- send
- add device 
- remove device
- create note
- redeem note

Those are all onchain ops. Pattern would fit future other ops (like waiting for approval from an onramp) as well.

- status `idle` > `loading` > `success`/`failure`
- at each point, a short updating message to show granular progress

i think we can use this hook for telemetry, to track anonymized / rolled-up performance and reliability data

apps/daimo-mobile/src/action/useCreateAccount.ts
```diff
@@ -0,0 +1,117 @@
...
+  }>();
+  useEffect(() => {
+    createEnclaveKey(setAS).then(setKey);
+  }, []);
+
+  // On exec, create contract onchain, claiming name.
+  const result = trpc.deployWallet.useMutation();
+  const exec = async () => {
+    if (!key?.pubKeyHex) return;
+    setAS("loading", "Deploying contract...");
```
nalinbhardwaj, line 28: I guess these logs/comments need to be updated (since no more deploy on this API?)

apps/scratchpad/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  const depositTxHash = await entrypoint.write.depositTo([addr], {
+    value: 10n ** 16n, // 0.01 ETH
+  });
+  console.log(`Faucet deposited prefund: ${depositTxHash}`);
+  await waitForTx(publicClient, depositTxHash);
+
+  // Finally, we should be able to do a userop from our new Daimo account.
+  // Send $0.50 USDC to nibnalin.eth
+  const recipient = `0xF05b5f04B7a77Ca549C0dE06beaF257f40C66FDB`;
+  const opHashApprove = await account.erc20transfer(recipient, "500000");
```
nalinbhardwaj, line 98: should this be `0.5`? don't we parse in @daimo/userop?

apps/daimo-mobile/src/view/screen/SendScreen.tsx
```diff
@@ -76,51 +79,116 @@ function Scan({ hide }: { hide: () => void }) {
...
+  const [account] = useAccount();
+  assert(account != null);
+
+  const { status, message, exec } = useSendTransaction(
+    account.enclaveKeyName,
+    recipient.account.addr,
+    dollars
+  );
+
+  // TODO: load estimated fees
```
nalinbhardwaj, line 144: ah this was the intention for the `dryRun` parameter in @daimo/userop but I just realised that still asks for a signature: should change that behaviour in a seperate PR and use that

nalinbhardwaj (approved):
🔥

apps/scratchpad/src/index.ts
dcposch: fixed

apps/daimo-mobile/src/action/useCreateAccount.ts
dcposch: not yet but will soon.

this endpoint should do a single, atomic userOp that
- materializes the account
- infinite approves USDC to the paymaster
- claims the name

PR #39, opened 2023-06-17
@daimo/contracts: add tokenMetadata

nalinbhardwaj:
Create an autogeneration script to fetch and populate common metadata about the ERC20 token. Saves us a fetch in app or hardcoding at places.


apps/daimo-mobile/src/view/screen/DepositScreen.tsx
```diff
@@ -92,7 +95,8 @@ function OnrampStub() {
     <View style={styles.callout}>
       <TextBody>
         <Octicons name="alert" size={16} color="black" />{" "}
-        <TextBold>Onramp coming soon.</TextBold> You'll be able to buy USDC
+        <TextBold>Onramp coming soon.</TextBold> You'll be able to buy $
+        {tokenMetadata.symbol}
```
dcposch, line 99: don't need the $, this will render as `$USDC`

nalinbhardwaj: good catch, fixed!

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
```diff
@@ -60,9 +61,9 @@ function IntroPages({ onCreateAccount }: { onCreateAccount: () => void }) {
             technology. Use at your own risk."
         />
         <IntroPage
-          title="USDC"
-          body="Daimo lets you send and receive money using the USDC stablecoin.
-            1 USDC is $1. Learn how it works here."
+          title={tokenMetadata.symbol}
```
dcposch, line 64: hmmmm.

if someone forks Daimo, they should write their own onboarding copy.

packages/daimo-contract/src/tokenMetadata.ts
```diff
@@ -0,0 +1,9 @@
+// Generated by script/fetchTokenMetadata.ts
+
+export const tokenMetadata = {
+  address: "0x1B85deDe8178E18CdE599B4C9d913534553C3dBf",
+  decimals: 6,
+  symbol: "USDC",
+  name: "testUSDC",
+  totalSupply: "1000000000000",
```
dcposch, line 8: tbh this is a tiny amount of metadata

for a huge decision (which coin Daimo runs on)

idk if codegen is necessary for that

(not a big deal though; checking it in to `daimo-contract` so we can import everywhere else lgreattm)

dcposch (approved):
LGTM

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
dcposch: fine here

might add a forking guide in future

packages/daimo-contract/src/tokenMetadata.ts
nalinbhardwaj: yeah, good point!

PR #40, opened 2023-06-18
Recover account from enclave after reinstall

dcposch:
If the user uninstalls the app, app storage is deleted but the enclave key is still there.

(It's not guaranteed to last any particular length of time.)

If user uninstalls+reinstalls the app (say, in response to flakiness--observed user habit, according to Moxie) this lets them recover their account.

### Done

Frontend recovery

### Not done

Backend is a stub. 

We have an indexing challenge:
- To lookup account by signing key cleanly, we need an `AddKey` event or onchain registry
- Onchain registry would be unacceptable cost on mainnet, but on L2 it might be near-free; you mainly pay for calldata (=L1 DA), so simply storing a `map(bytes32 => address)` from signing key hash to account might work.


dcposch:
Leaving the backend for later.

Learned two things:
- We need better tracking for signing keys. See above.
- Enclave keys do, in fact, survive uninstall. Will need further testing later to determine how reliably and for how long.

nalinbhardwaj:
> If the user uninstalls the app, app storage is deleted but the enclave key is still there.
>
> (It's not guaranteed to last any particular length of time.)

At least on iOS, private key bytes are written to a [keychain item](https://developer.apple.com/forums/thread/36442) which online sources suggest are only deleted on device wipe/manual deletion by user/app (regardless of length of time). Empirically, my laptop has a keychain item from an app I deleted in 2019.

dcposch:
@nalinbhardwaj it currently seems to last indefinitely, but not guaranteed to stay that way: https://developer.apple.com/forums/thread/36442?answerId=281900022#281900022

PR #46, opened 2023-06-19
Websocket transport

dcposch:
## Use websocket transport

This should speed up onchain lookups.

## Do we need local chain RPC at all?

An alternative would be to route everything through the Daimo API.

- Make the Daimo API pluggable and (eventually) trust-minimized
- Connect via websocket for efficient event subscriptions
- Make it relatively easy for other parties to run performant API servers
- Stretch goal: app ships with at least two independent API endpoints. Advanced setting to add or change. App tracks performance of each endpoint, sends traffic to the ones that are available & fastest. This serves both performance and censorship resistance.
- End goal: app ships with a light client (probably `helios`). The API server is temporarily trusted for fast transaction confirmation, but verified via light client proofs.

### DIFF
--- a/apps/daimo-mobile/package.json
+++ b/apps/daimo-mobile/package.json
@@ -9,7 +9,6 @@
     "build:stage": "eas build -p ios --profile=stage",
     "submit:stage": "eas submit --platform ios --profile=stage",
     "start": "expo start --dev-client --clear -i",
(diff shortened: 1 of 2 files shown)


dcposch:
Closing; app will use just the API. API will be trust-minimized via light client and federated.

PR #47, opened 2023-06-20
Deep link to request payment

dcposch:


dcposch:
Superseded by #95 

We don't need Branch. We need universal deep links, not deferred links.

For the future, there are three levels of app link aka deep link:

- Bare links: `daimo://request/...`. These work only if the recipient has the app already installed, otherwise they go nowhere. They also show as bare text or links--no nice preview in messenger apps.
- Universal links: `https://daimo.com/request/...`. These can show a nice preview. If the user has the app installed, it takes them directly to the app. If not, they see a website that can direct to the app store. Requires AASA "apple app site association", see [this guide](https://docs.expo.dev/guides/deep-linking/).
- Deferred deep links: same as above, but after installing the app it automatically opens the requested action (without them having to click the link again). This requires fancy hacks as implemented by Branch.

PR #53, opened 2023-06-23
Faster send

dcposch:
Cache DaimoAccount init

## Before

~1500ms to signing

## After

```
 LOG  [ACTION] 0ms: loading Loading account...
 LOG  [ACTION] 24ms: loading Signing...
 ```

### DIFF
--- a/packages/daimo-api/src/contract/nameRegistry.ts
+++ b/packages/daimo-api/src/contract/nameRegistry.ts
@@ -17,10 +17,12 @@ import {
 import { ClientsType, ContractType, getClients } from "../chain";
 import { NamedAccount } from "../model";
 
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: packages/daimo-api/src/contract/nameRegistry.ts [hunk 1]
```diff
@@ -99,6 +104,10 @@ export class NameRegistry {
       hash,
     });
     if (tx.status !== "success") throw new Error("Transaction failed");
+
+    // Cache
+    this.cacheAccount({ name, addr: address });
```
dcposch, line 109: fixes a bug where newly created accounts wouldn't show up in search until API restart

PR #56, opened 2023-06-27
add/remove device

nalinbhardwaj:


apps/daimo-mobile/src/logic/device.ts
```diff
@@ -0,0 +1,53 @@
...
+  "🥕",
+  "🌽",
+  "🌶️",
+  "🍄",
+  "🥒",
+  "🥦",
+  "🍞",
+  "🥖",
+  "🥐",
+];
```
dcposch, line 36: hmm given that there's a canonical list of keys

we could report the position in the list, maybe Device A / B / C

apps/daimo-mobile/src/logic/device.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function pubKeyToEmoji(pubKey: Hex): string {
+  return emojis[keccak256(pubKey, "bytes")[0] % emojis.length];
```
dcposch, line 39: tough thing with this approach:
- will be fairly common for people to have two `Device 🍄`s
- emojis are distinct style. its worth considering, but if we do it we'll likely end up using emojis elsewhere in the UI too

apps/daimo-mobile/src/model/account.ts
```diff
@@ -41,46 +44,67 @@ export type Account = {
...
+//   lastBalance: string;
+//   lastFinalizedBlock: number;
+//   recentTransfers: TransferOpEvent[];
+//   trackedRequests: TrackedRequest[];
+//   namedAccounts: EAccount[];
 
-interface AccountV4 extends StoredModel {
-  storageVersion: 4;
+//   pushToken: string | null;
+// }
```
dcposch, line 101: delete commented ^

apps/daimo-mobile/src/model/account.ts
```diff
@@ -171,54 +196,15 @@ export function parseAccount(accountJSON?: string): Account | null {
...
-      lastFinalizedBlock: a.lastFinalizedBlock,
-
-      recentTransfers: a.recentTransfers,
-      trackedRequests: [],
-      namedAccounts: a.namedAccounts,
-
-      pushToken: a.pushToken,
-    };
-  }
+  // Delete V1-4 testnet accounts. Re-onboard to accounts with multi-key support.
```
dcposch, line 199: was skeptical, but i'm convinced now that it's good if & only if we ship concurrently with a much faster onboarding experience.

gives our testers a nice surprise

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -257,13 +259,3 @@ function addTransfers(
 
   return ret;
 }
-
-/** On L2, timestamp can be a deterministic function of block number. */
```
dcposch, line 261: 👍 👍

apps/daimo-mobile/src/view/screen/AddDeviceScreen.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const [barCodeStatus, setBarCodeStatus] = useState<
+    "idle" | "error" | "scanned"
+  >("idle");
+
+  const handleBarCodeScanned: BarCodeScannedCallback = async ({ data }) => {
+    if (barCodeStatus !== "idle") return;
+
+    try {
+      const pubkeyHex = parseAddKeyString(data);
+      setBarCodeStatus("scanned");
```
dcposch, line 42: no need for this PR, but in future we'll likely need a way to paste in a new device key as an alternative to QR scanning

apps/daimo-mobile/src/view/screen/AddDeviceScreen.tsx
```diff
@@ -0,0 +1,131 @@
...
+      )}
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  vertOuter: {
+    flex: 1,
+    padding: 32,
+    gap: 16,
```
dcposch, line 128: nit but i recommend `<Spacer height={16} />`s,  makes it explicit and adjustable

apps/daimo-mobile/src/view/screen/DeviceScreen.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setAccount(null);
+        nav.navigate("Home");
+      }
+    }
+  }, []);
+
+  const statusMessage = (function (): ReactNode {
+    switch (status) {
+      case "idle":
+        return `Remove fee: ${getAmountText({ dollars: cost.totalDollars })}`;
```
dcposch, line 101: just `Fee: `

apps/daimo-mobile/src/view/screen/DeviceScreen.tsx
```diff
@@ -0,0 +1,160 @@
...
+          <Spacer h={16} />
+        </>
+      )}
+      <TextBody>
+        <TextCenter>
+          Added at {timeString(guessTimestampFromNum(device.addedAt))}
+        </TextCenter>
+      </TextBody>
+      <Spacer h={16} />
+      <TextCenter>Last used at unknown</TextCenter>
```
dcposch, line 149: intentional?

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
```diff
@@ -37,36 +35,37 @@ import {
...
-
+  // TODO: add back buttons?
   return (
     <View style={styles.onboardingScreen}>
-      {page === 1 && <IntroPages onCreateAccount={next} />}
+      {page === 1 && <IntroPages onFlowChoice={onFlowChoice} />}
       {page === 2 && <AllowNotifications onNext={next} />}
-      {page === 3 && <CreateAccountPage />}
+      {accountFlowChoice === "create" && page === 3 && <CreateAccountPage />}
+      {accountFlowChoice === "existing" && page === 3 && <UseExistingPage />}
```
dcposch, line 59: 👍 👍

apps/daimo-mobile/src/view/screen/SettingsScreen.tsx
```diff
@@ -73,61 +76,101 @@ export function SettingsScreen() {
...
+    </ScrollView>
+  );
+}
+
+function DevicesInfo({ account }: { account: Account }) {
+  const nav = useNav();
+
+  const addDevice = () => nav.navigate("AddDevice");
+
+  console.log(`[NALIN DEBUG] accountKeys`, JSON.stringify(account.accountKeys));
```
dcposch, line 122: delete

apps/daimo-mobile/src/view/screen/SettingsScreen.tsx
```diff
@@ -73,61 +76,101 @@ export function SettingsScreen() {
...
+  const nav = useNav();
+
+  const viewDevice = () => nav.navigate("Device", { pubKey: keyData.key });
+
+  return (
+    <ButtonSmall onPress={viewDevice}>
+      <View style={styles.deviceDataRow}>
+        <TextBold>
+          Device {pubKeyToEmoji(keyData.key)}
+          {isCurrentDevice ? " (Current device)" : " "}
```
dcposch, line 164: 👍

dcposch (commented):
Looks excellent. Left comments. Reviewing the contracts next.

packages/contract/src/Account.sol
```diff
@@ -163,4 +163,45 @@ contract Account is BaseAccount, UUPSUpgradeable, Initializable {
...
+                accountKeys[i][1] == accountPubKey[1]
+            ) {
+                for (uint256 j = i;j < accountKeys.length - 1; j++) {
+                    accountKeys[j] = accountKeys[j + 1];
+                }
+                accountKeys.pop();
+                emit SigningKeyRemoved(this, accountPubKey);
+                return;
+            }
+        }
```
dcposch, line 204: looks great

one alternative to consider

- addSigningKey, removeSigningKey both take (index, pubKey)
- addSigningKey asserts `accountKeys[index]` is 0, then sets to pubKey
- removeSigningKey asserts `accountKeys[index]` equals pubKey, then sets to 0
- either way, assert index is in range 0-255
- track numSigningKeys, removeSigningKey asserts `numSigningKeys > 1`

advantages:
- both are now constant time/constant gas
- lets us refer to Device A/B/C or 1/2/3 without "Device C" changing into "Device B" when you remove B
- fixes an edge-case: if Device C has bad or no internet & doesn't yet know that Device B has been removed, and you sign an offline transaction from Device C (see Vitalik's "personal experiences" post for scenario), then that transaction will be still be valid (random nonce + stable keyIndex allows this to work)

dcposch: i recommend adding `(index, pubKey)` to the `SigningKeyAdded/Removed` event too

apps/daimo-mobile/src/logic/device.ts
nalinbhardwaj: If devices are removed and added the positions shift which might create confusion.

apps/daimo-mobile/src/logic/device.ts
nalinbhardwaj: Yeah, I agree, makes sense! Just wanted a stand-in for labelling for now. Given that we want to represent ~256 bits of entropy we should think about a better idea in general.

apps/daimo-mobile/src/view/screen/DeviceScreen.tsx
nalinbhardwaj: yeah stub to show we'll add it future but don't feel strongly about keeping it, also changed it to `Last used at ??? (TODO)` so it doesnt look like a bug.

packages/daimo-api/src/contract/keyRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+    const currentBlockNumber = await this.vc.publicClient.getBlockNumber(); // TODO: remove?
+
+    const addrToNewLogs: Map<Address, SigningKeyAddedOrRemovedLog[]> =
+      new Map();
+    for (const log of logs) {
+      const addr = getAddress(log.address);
+      if (addrToNewLogs.get(addr) === undefined) {
+        addrToNewLogs.set(addr, []);
+      }
+      addrToNewLogs.get(addr)!.push(log);
```
dcposch, line 86: you can write

```
const newLogs = addrToNewLogs.get(addr) || [];
addrToNewLogs.set(addr, newLogs.concat([log]));
```

packages/daimo-api/src/contract/keyRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+        `[API] Failed to get keys for ${addr}, probably an old account?`,
+        e
+      );
+      return [];
+    }
+  }
+
+  /** Parses account key add/remove logs, first on init() and then on subscription. */
+  parseLogs = async (logs: SigningKeyAddedOrRemovedLog[]) => {
+    const currentBlockNumber = await this.vc.publicClient.getBlockNumber(); // TODO: remove?
```
dcposch, line 77: i'd either remove this, or pass it into `getKeys` to ensure we're fetching keys as of that block number

packages/daimo-api/src/contract/keyRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+  /** Cache an address's key properties in memory. */
+  cacheAddressProperties = (
+    addr: Address,
+    currentKeySet: Hex[],
+    currentBlockNumber: bigint
+  ) => {
+    // deterministically sort all logs
+    const sortedLogs = this.addrToLogs.get(addr)!.sort((a, b) => {
+      const aBlockNumber = a.blockNumber || currentBlockNumber + 1n;
+      const bBlockNumber = b.blockNumber || currentBlockNumber + 1n;
```
dcposch, line 113: when does a log not have `blockNumber`?

dcposch: i think we can `assert(a.blockNumber != null && a.logIndex != null)`

dcposch: in which case this reduces to:

```ts
let diff = a.blockNumber - b.blockNumber;
if (diff !== 0) return diff;
return a.logIndex - b.logIndex;
```

packages/daimo-api/src/contract/nameRegistry.ts
```diff
@@ -46,7 +46,7 @@ export class NameRegistry {
   }
 
   /** Parses Registered event logs, first in init(), then on subscription. */
-  parseLogs = (logs: RegisteredLog[]) => {
+  parseLogs = async (logs: RegisteredLog[]) => {
```
dcposch, line 49: this isn't async, right?

packages/daimo-api/src/pushNotifier.ts
```diff
@@ -20,6 +21,7 @@ export class PushNotifier {
   constructor(
     private coinIndexer: CoinIndexer,
     private nameReg: NameRegistry,
+    private keyReg: KeyRegistry, // TODO: notify devices on key add/remove
```
dcposch, line 24: 👍

packages/daimo-common/src/key.ts
```diff
@@ -0,0 +1,33 @@
...
+const derPrefix = "0x3059301306072a8648ce3d020106082a8648ce3d03010703420004";
+
+export function isDERPubKey(pubKeyHex: Hex): boolean {
+  return (
+    pubKeyHex.startsWith(derPrefix) &&
+    pubKeyHex.length === derPrefix.length + 128
+  );
+}
+
+export function DERKeytoContractFriendlyKey(pubKeyHex: Hex): [Hex, Hex] {
```
dcposch, line 14: nit, `derKeyToContractFriendlyKey`

i recommend adding a unit test for `key.ts`

packages/daimo-common/src/model.ts
```diff
@@ -78,3 +78,13 @@ export const zTrackedRequest = z.object({
 });
 
 export type TrackedRequest = z.infer<typeof zTrackedRequest>;
+
+export const zKeyData = z.object({
+  key: zHex, // DER Format
+  addedAt: z.number(),
+  removedAt: z.number().optional(),
+  keyIndex: z.number().optional(),
```
dcposch, line 86: can be mandatory now

+ `lastUsedAt: z.number().optional()` lgtm

packages/daimo-api/src/contract/keyRegistry.ts
nalinbhardwaj: done, iirc at the time viem had a weird types bug (that's the TODO next to `currentBlockNumber` as well), that seems fixed now so

apps/daimo-mobile/src/view/screen/DeviceScreen.tsx
dcposch: can also comment for now `{/* <TextCenter>Last used at unknown</TextCenter> */}`

nalinbhardwaj: done

packages/daimo-common/src/key.ts
nalinbhardwaj: done both.

apps/daimo-mobile/test/device.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it("finds unused slot", () => {
+    expect(findUnusedSlot([])).toEqual(0);
+    expect(findUnusedSlot([0])).toEqual(1);
+    expect(findUnusedSlot([0, 1])).toEqual(2);
+    expect(findUnusedSlot([0, 2])).toEqual(3);
+    expect(findUnusedSlot([0, 255])).toEqual(1);
+    expect(findUnusedSlot([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])).toEqual(10);
+    expect(findUnusedSlot([0, 1, 2, 4, 5, 6, 7, 8, 9, 255])).toEqual(3);
```
dcposch, line 25: 🚀

apps/daimo-mobile/src/action/key.ts
```diff
@@ -10,7 +10,7 @@ import { defaultEnclaveKeyName } from "../model/account";
 function getKeySecurityMessage(hwSecLevel: ExpoEnclave.HardwareSecurityLevel) {
   switch (hwSecLevel) {
     case "SOFTWARE":
-      return "🔒  Key generated in Secure Enclave";
+      return "🔒  Key generated in software";
```
dcposch, line 13: minor but i'd either remove the key logo here or add it on `Key generated in trusted hardware` below for consistency

packages/contract/src/Account.sol
```diff
@@ -14,23 +14,22 @@ import "account-abstraction/core/BaseAccount.sol";
...
 
     IEntryPoint private immutable _entryPoint;
 
     P256SHA256 private immutable _sigVerifier;
 
     event AccountInitialized(IEntryPoint indexed entryPoint);
-    event SigningKeyAdded(IAccount indexed account, bytes32[2] accountPubkey);
-    event SigningKeyRemoved(IAccount indexed account, bytes32[2] accountPubkey);
+    event SigningKeyAdded(IAccount indexed account, uint8 keySlot, bytes32[2] key);
+    event SigningKeyRemoved(IAccount indexed account, uint8 keySlot, bytes32[2] key);
```
dcposch, line 32: 🚀 🚀

dcposch (approved):
LGreatTM!

packages/contract/src/Account.sol
nalinbhardwaj: done both

packages/daimo-common/src/model.ts
nalinbhardwaj: done

apps/daimo-mobile/src/action/key.ts
nalinbhardwaj: removed, makes sense

apps/daimo-mobile/src/model/account.ts
nalinbhardwaj: done

apps/daimo-mobile/src/logic/device.ts
nalinbhardwaj: replaced by slots/A/B/C now

apps/daimo-mobile/src/logic/device.ts
nalinbhardwaj: replaced by slots/A/B/C now

apps/daimo-mobile/src/view/screen/AddDeviceScreen.tsx
nalinbhardwaj: done

packages/daimo-api/src/contract/keyRegistry.ts
nalinbhardwaj: done

apps/daimo-mobile/src/view/screen/DeviceScreen.tsx
nalinbhardwaj: done

PR #79, opened 2023-07-09
Update README.md

nalinbhardwaj:


dcposch (approved):
LGTM. feel free to merge updates like this directly

(we'll get more careful over time. right now we're in rapid prototyping mode.)

PR #98, opened 2023-07-18
Recipient Confirmation Codes

onsen-egg:
This is only a high-level feature proposal and not a functioning branch atm; I just wanted to include a bit of code as a rough illustration.

## Summary
When you send a Venmo payment to an unknown recipient, you get prompted to enter the last 4 digits of their phone number as an optional way of confirming it's the right person. A simple way to provide a similar experience (if paired with the right UX) would be to define a recipient's "confirmation code" to be the last 4 or so hex digits of the hash of their wallet address, and use it in the same way.

You could have the app frontend just derive the code on the fly from the recipient address that the user inputs, but then using it would be completely contrived and wouldn't prevent any mistakes. Thus, the way I imagine UX working is that in the frontend, you'd omit the logic for deriving confirmation codes for anyone except the user's own wallet address, which they could then share along with their address/ENS name when receiving a payment.

Of course, doing this check on a tx locally before prompting for a signature, rather than afterwards on-chain, would probably be just as good, but thanks to account abstraction this is pretty painless to implement on the smart contract side—I wrote out a quick sketch of what that would look like.

While ENS domains and QR codes are definitely the most important things, I would personally quite like having a light optional guardrail like this as well, and I think it would help prevent mistakes even in a world with ubiquitous ENS usage (w.r.t. domain name lookalikes, for example). For me it takes away the anxiety when I'm entering a recipient manually—it's a way to feel basically certain that I've entered the recipient correctly that isn't just looking really closely at what I've just entered, which imo isn't the greatest, even when the addresses are human-readable.

Keen for some feedback on this idea—if you all were on board, I might take some time to prototype a frontend integration, or move it over there entirely.

### DIFF
--- a/packages/contract/script/DemoExec.s.sol
+++ b/packages/contract/script/DemoExec.s.sol
@@ -18,6 +18,7 @@ contract DemoExecScript is Script {
         newAcc.execute(
             0xF05b5f04B7a77Ca549C0dE06beaF257f40C66FDB,
             0.01 ether,
+            0x0,
             data
         );
 
--- a/packages/contract/src/Account.sol
+++ b/packages/contract/src/Account.sol
@@ -63,8 +63,13 @@ contract Account is BaseAccount, UUPSUpgradeable, Initializable {
     function execute(
         address dest,
         uint256 value,
+        bytes2 confirmationCode,
         bytes calldata func
     ) external {
+        if (confirmationCode != 0) {
+            _validateConfirmationCode(confirmationCode, func);
+        }
+
         _requireFromEntryPoint();
         _call(dest, value, func);
     }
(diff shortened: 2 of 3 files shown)


dcposch:
Thanks for the writeup!

We're at ethCC this week. A lot going on, so took me longer to see this PR.

A few thoughts:
- Making sure you're paying the right person is a key challenge, especially for a noncustodial wallet.
- 4-digit code is an interesting idea, but I'm not sure how much it helps here. Venmo benefits from the fact that people already have each other's numbers in their phone, and can often double-check the recipient that way.
- If we do ask the user to enter a code derived from the recipient address, I don't think it needs to be in the contract. Instead, it can be a UI-level check.


Curious, how'd you hear about Daimo?

onsen-egg:
- Compared with Venmo, I def agree it doesn't provide the same guarantees—it doesn't help in the adversarial case, where someone is impersonating a recipient you know and giving you the wrong address in the first place. But if the right person is trying to communicate the right address, and you're entering it manually, it saves you from typos or similar-looking/sounding but wrong addresses/ENS names.
- Definitely agree that this would only have to exist at the UI level

Saw a tweet from Nalin about it, thought it was a cool project :)

onsen-egg:
Come to think of it though, what if you had the option to post a hiding commitment to some piece of personal information, which your friends could use to confirm when sending you a payment? So as a concrete example, I could:
- Take the hash of the last four digits of my phone number + my wallet address, then put that hash on chain using the same address (the address in the hash is just acting as a salt to prevent re-use)
- When someone who knows me is paying me, they can get prompted for those four digits, and can confirm that it matches that address and the hash I've associated it with. Since this only happens in the UI the information never gets revealed on-chain

This could of course generalize to any personal info or user-chosen code as long as there's some standard schema in place.

dcposch:
I appreciate the idea.

I think this is more a feature request than PR. Diff adds a `confirmationCode` argument to our onchain 4337 account... if we end up implementing a confirmation PIN, like Venmo, it will be UI-only.

May revisit. Gotta ship a working app first.

PR #105, opened 2023-07-21
OP: Change nonce logic to include metadata

### LINKED ISSUE #100: Remove daimo-userop RPC dependency
Currently, `DaimoAccount.init` takes a a Viem publicClient.

This is used for exactly one purpose: pulling the latest `nonce` from the entryPoint contract.

Removing this dependency means that `daimo-userop` only talks to a bundler RPC, instead of both a bundler RPC and an ethereum JSON RPC.

Means that Daimo overall has two server dependencies--bundler and API--instead of 3. In future, we can reduce this to one by routing everything thru the API and trust-minimizing via light client.

nalinbhardwaj:
Closes #100

Depends on #160.

### DIFF
--- a/apps/daimo-mobile/src/action/useCreateAccount.ts
+++ b/apps/daimo-mobile/src/action/useCreateAccount.ts
@@ -60,7 +60,6 @@ export function useCreateAccount(name: string): ActHandle {
       name,
       address,
 
-      isDeployed: false,
       lastBalance: BigInt(0),
       lastBlockTimestamp: 0,
       lastBlock: 0,
--- a/apps/daimo-mobile/src/action/useCreateAccount.ts
+++ b/apps/daimo-mobile/src/action/useCreateAccount.ts
@@ -24,7 +24,7 @@ export function useCreateAccount(name: string): ActHandle {
   const result = rpcHook.deployWallet.useMutation();
   const exec = async () => {
     if (!pubKeyHex) return;
-    setAS("loading", "Registering name...");
+    setAS("loading", "Deploying account...");
     result.mutate({ name, pubKeyHex });
   };
 
--- a/apps/daimo-mobile/src/action/useCreateAccount.ts
+++ b/apps/daimo-mobile/src/action/useCreateAccount.ts
@@ -24,7 +24,7 @@ export function useCreateAccount(name: string): ActHandle {
   const result = rpcHook.deployWallet.useMutation();
   const exec = async () => {
     if (!pubKeyHex) return;
-    setAS("loading", "Deploying contract...");
+    setAS("loading", "Registering name...");
     result.mutate({ name, pubKeyHex });
   };
 
@@ -60,6 +60,7 @@ export function useCreateAccount(name: string): ActHandle {
       name,
       address,
 
+      isDeployed: false,
       lastBalance: BigInt(0),
       lastBlockTimestamp: 0,
       lastBlock: 0,
--- a/apps/daimo-mobile/src/model/account.ts
+++ b/apps/daimo-mobile/src/model/account.ts
@@ -1,4 +1,5 @@
 import { EAccount, assert, TransferOpEvent } from "@daimo/common";
+import { ephemeralNotesAddress } from "@daimo/contract";
 import { useEffect, useState } from "react";
 import { MMKV } from "react-native-mmkv";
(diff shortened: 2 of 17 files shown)


apps/daimo-mobile/src/model/account.ts
```diff
@@ -150,6 +171,7 @@ export function parse(accountJSON?: string): Account | null {
       name: a.name,
       address: getAddress(a.address),
 
+      isDeployed: true, // TODO: is this okay to assume?
```
nalinbhardwaj, line 174: this migration path will _probably_ be okay given current TestFlight set, but in future would be nice to be able to sequentially make migrations and have tests for different version storage

apps/daimo-mobile/src/model/account.ts
```diff
@@ -162,13 +184,41 @@ export function parse(accountJSON?: string): Account | null {
...
 
-  assert(model.storageVersion === 3);
-  const a = model as AccountV3;
+  if (model.storageVersion === 3) {
+    const a = model as AccountV3;
+
+    const isDeployed =
+      a.recentTransfers.find(
+        (t) =>
+          t.from === getAddress(a.address) || t.from === ephemeralNotesAddress
```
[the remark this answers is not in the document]
dcposch: I recommend adding `isDeployed` to `getAccountHistory` (the sync endpoint) in `router.ts`

the migration can always initialize to `false` when unknown

& on first sync, will update to correct value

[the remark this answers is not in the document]
nalinbhardwaj: Done, extra RPC call in the sync but definitely cleaner.

apps/daimo-mobile/src/model/account.ts
```diff
@@ -70,6 +72,24 @@ interface AccountV3 extends StoredModel {
   pushToken: string | null;
 }
 
+interface AccountV4 extends StoredModel {
```
dcposch, line 75: we might not need an `Account` migration

`isDeployed` should always be true. we send the first transaction at account creation time, claiming the name

so by the time we call `setAccount` to save to storage, it's already deployed

nalinbhardwaj: > we send the first transaction at account creation time, claiming the name

I don't think so? Name claim just assigns the name to the counter factual address, but doesn't deploy it atm I believe?

nalinbhardwaj: Example, https://goerli.basescan.org/tx/0x13d5ea27049dc4f84d94b3ba0616658675a2e75de4b212c93d0a308f9712c6db is deployed at first transfer.

dcposch: you're right. sending the tx from the faucet address was a shortcut, i think i have a TODO in there...

given we're spotting gas for the name claim, we can send it from the account (currently paying gas via prefund, another shortcut, soon via paymaster) which will instantiate it

if we don't do that, the first outbound transfer will have higher than normal fees

simplifying the stored account is a bonus

nalinbhardwaj: > given we're spotting gas for the name claim, we can send it from the account which will instantiate it

hm, looking more closely at this right now and i think we should still send the tx from our faucet/api backend address.

Benefits:
- Saves a step in user onboarding flow (biometrics for deploy signature).
- If the server deploys and returns the deployed address before we initialize the userop package, it doesn't need to care about calling `initCode`/`getSenderAddress` and we can get rid of the RPC call there. Server is trusted but that's fine I think.

nalinbhardwaj: Marking resolved: done in #160 and rebased this on top of that.

apps/daimo-mobile/src/action/useCreateAccount.ts
```diff
@@ -24,7 +24,7 @@ export function useCreateAccount(name: string): ActHandle {
   const result = rpcHook.deployWallet.useMutation();
   const exec = async () => {
     if (!pubKeyHex) return;
-    setAS("loading", "Deploying contract...");
```
dcposch, line 27: this is user facing, so i'd still call it an account

related: "the user doesn't know what a key is" from this great guide

https://github.com/signalapp/Signal-Android/blob/main/CONTRIBUTING.md

PR #160, opened 2023-08-16
Move deploy from first userop to backend

nalinbhardwaj:
Moves deployment of wallet from first userop to backend, as discussed in https://github.com/daimo-eth/daimo/pull/105#discussion_r1294798932, this makes reasoning about account state much cleaner after onboarding and will also allow us to remove RPC dependency from `@daimo/userop`.

Depends on https://github.com/stackup-wallet/userop.js/pull/95, don't merge until that dependency is replaced.


packages/daimo-api/src/router.ts
```diff
@@ -272,28 +273,33 @@ export function createRouter(
...
-          () => {
-            throw new Error("No signer");
-          },
-          false
-        );
-        const address = account.getAddress();
+        const address = await accountFactory.getAddress(pubKeyHex);
+
+        // TODO: Should be able to batch these in a single tx for perf.
+        // Promise.all awaiting doesn't work because nonces go out of sync.
```
nalinbhardwaj, line 279: Waiting on these three txes makes onboarding quite slow, we should do this asap a follow up PR. Ironically it seems the "right way" to do this is to use an AA wallet for our backend as well lol.

apps/daimo-mobile/src/action/useSendAsync.ts
```diff
@@ -75,7 +76,7 @@ function loadAccount(enclaveKeyName: string) {
       chain: chainConfig.l2,
       transport: http(),
     });
-    return await DaimoAccount.init(client, derPublicKey, signer, false);
+    return await DaimoAccount.init(client, address, signer, false);
```
dcposch, line 79: 👍

apps/daimo-mobile/src/action/useSendAsync.ts
```diff
@@ -85,11 +86,12 @@ function loadAccount(enclaveKeyName: string) {
 async function sendAsync(
   setAS: SetActStatus,
   enclaveKeyName: string,
+  address: Address,
   sendFn: SendOpFn
 ) {
   try {
     setAS("loading", "Loading account...");
-    const account = await loadAccount(enclaveKeyName);
+    const account = await loadAccount(enclaveKeyName, address);
```
dcposch, line 94: 👍

apps/scratchpad/src/createAccount.ts
```diff
@@ -64,24 +69,52 @@ export async function createAccount() {
...
+  const key1 = `0x${pubKey.subarray(0, 32).toString("hex")}` as Hex;
+  const key2 = `0x${pubKey.subarray(32).toString("hex")}` as Hex;
+  const salt = 0n;
+  const args = [[key1, key2], salt] as const;
+
+  const address = await publicClient.readContract({
+    ...accountFactoryConfig,
+    functionName: "getAddress",
+    args,
+  });
```
dcposch, line 87: step 1: call RPC to get our own counterfactual address

(we could compute this ourselves. not urgent but a nice future optimization)

apps/scratchpad/src/createAccount.ts
```diff
@@ -64,24 +69,52 @@ export async function createAccount() {
...
+    chain,
+    transport: http(),
+  });
+
+  // Deploy account
+  const hash = await walletClient.writeContract({
+    ...accountFactoryConfig,
+    functionName: "createAccount",
+    args,
+  });
```
dcposch, line 103: step 2: createAccount call, **wait several seconds for tx inclusion**

apps/scratchpad/src/createAccount.ts
```diff
@@ -64,24 +69,52 @@ export async function createAccount() {
...
   // Fund it from the faucet
-  const faucetPrivKey = `0x${process.env.DAIMO_API_PRIVATE_KEY}` as const;
-  const faucetAccount = privateKeyToAccount(faucetPrivKey);
-  console.log(`Faucet account: ${faucetAccount.address}`);
-  const walletClient = createWalletClient({
-    account: faucetAccount,
-    chain,
-    transport: http(),
-  });
   const usdcTxHash = await walletClient.writeContract({
```
dcposch, line 118: step 3 and 4: send USDC to account, prefund EntryPoint, each with a sequential delay of a few seconds for inclusion

fine for the scratchpad. i agree re: doing it in a single transaction in-app

packages/daimo-api/src/contract/accountFactory.ts
```diff
@@ -0,0 +1,65 @@
+import { accountFactoryConfig } from "@daimo/contract";
+import { Hex, TransactionReceipt } from "viem";
+
+import { ViemClient } from "../chain";
+
+/* Interface to the AccountFactory contract. Creates Daimo accounts. */
+export class AccountFactory {
```
dcposch, line 7: i think we're not using this file anywhere

packages/daimo-api/src/router.ts
```diff
@@ -272,28 +273,33 @@ export function createRouter(
...
+        if (deployReceipt.status !== "success") {
+          return { status: deployReceipt.status, address: undefined };
+        }
+        if (registerReceipt.status !== "success") {
+          return { status: registerReceipt.status, address: undefined };
+        }
+        if (prefundReceipt.status !== "success") {
+          return { status: prefundReceipt.status, address: undefined };
+        }
+        return { status: "success", address };
```
dcposch, line 302: 🚀

packages/daimo-api/src/router.ts
dcposch: three other options:
1. **quick and dirty concurrency** - just kick off all 3 transactions in parallel, instead of one after the other. `await Promise.all` on the three transactions, then again for the three receipts.
1. **AccountFactory.deployAndClaimName** - a single function in AccountFactory that deploys and account, claims a name, and uses `value` (if passed, testnet only, unnecessary when we have a paymaster setup) to prefund with the EntryPoint
2. **deploy from account itself** - DaimoAccount initialized with a name, registers itself with the NameRegistry in first transaction. a few ways to get this transaction onchain, but the simplest would be to bundle it ourselves

dcposch (approved):
Looks great to me

packages/daimo-api/src/contract/accountFactory.ts
nalinbhardwaj: the new version is used for deployment and account factory related calls in the API: https://github.com/daimo-eth/daimo/pull/160/files/50df9851ad8bcc51b33ead873d93320af33b3948#diff-d47c5fcd14c4963fdceda8b0a2628726e1412787b1af4a426cecfd14b50bb518R282

nalinbhardwaj:
Merging with #105 as the two of them together are a sufficient workaround to the upstream https://github.com/stackup-wallet/userop.js/pull/95 as they haven't merged that yet.

PR #162, opened 2023-08-16
Fix depcheck Next config fail

nalinbhardwaj:
Fixes our instance of https://github.com/depcheck/depcheck/issues/812, I don't like that we have to change our code to make our linter tooling work correctly in general.

Before:
image

After:
image


nalinbhardwaj:
Going to merge rn to stop cascading these PRs and diff is simple, ptal post-commit @dcposch

PR #174, opened 2023-08-19
show Payment Link status

dcposch:
ds

### DIFF
--- a/apps/daimo-mobile/src/logic/note.ts
+++ b/apps/daimo-mobile/src/logic/note.ts
@@ -42,6 +42,6 @@ export function useEphemeralSignature(
   return ephemeralSignature;
 }
 
-// TODO: remove after we migrate to EIP-7212 contract
+// TODO: remove once EphemeralNote contract no longer requires it.
 // Dummy signature of correct length and valid r/s values (so ECDSA.sol doesn't revert)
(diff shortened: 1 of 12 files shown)


apps/daimo-mobile/src/view/screen/send/CreateNoteTab.tsx
```diff
@@ -37,35 +29,20 @@ import { TextCenter, TextError, TextLight } from "../../shared/text";
...
-      const allowance = await fetchNotesContractAllowance(address);
-      setIsNotesContractApproved(allowance > 0);
 
-      const privKey = generatePrivateKey();
-      setEphemeralPrivateKey(privKey);
-    })();
-  }, []);
+  // TODO: load and cache
+  // Minor optimization. Approving unconditionally costs a bit more gas.
+  const approveFirst = true;
```
nalinbhardwaj, line 45: I think we should consider moving this approve to the onboarding deployment step: would make `@daimo/userop` slightly cleaner as well

nalinbhardwaj (approved):
🚀

apps/daimo-mobile/src/logic/note.ts
```diff
@@ -115,5 +42,6 @@ export function useEphemeralSignature(
   return ephemeralSignature;
 }
 
+// TODO: remove after we migrate to EIP-7212 contract
```
nalinbhardwaj, line 45: This is independent of the P256 verifier change, this is a valid dummy signature for [ECDSA.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol)

Issue #191, opened 2023-08-24
Android "ExpoEnclave: unable to read account private key"


dcposch:
<img width="250" alt="image" src="https://github.com/daimo-eth/daimo/assets/169280/e5de736e-4c78-41aa-b88c-ae9c49e2c4b2"> <img width="250" alt="image" src="https://github.com/daimo-eth/daimo/assets/169280/a908bafc-a53a-4be6-a8dc-6682b51f641a">

dcposch:
Closing, can reopen if it reoccurs

PR #194, opened 2023-08-24
expo-enclave: Don't throw on failed reads

nalinbhardwaj:
Unclear why #191 happens since we're explicitly catching those throws (possible @dcposch had an old module build somehow?), this PR should clean up that throw logic anyway.


dcposch (approved):
LGTM

PR #205, opened 2023-08-28
Polling-based log sync

dcposch:
Plus local file cache for fast restart

before: several minutes to resync `getLogs` from scratch
after: 12 seconds

### DIFF
--- a/packages/daimo-api/src/chain.ts
+++ b/packages/daimo-api/src/chain.ts
@@ -1,9 +1,12 @@
 import type { AbiEvent } from "abitype";
+import jsonBigInt from "json-bigint";
+import fs from "node:fs/promises";
+import os from "os";
+import path from "path";
 import {
   Abi,
   Account,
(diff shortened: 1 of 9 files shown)


packages/daimo-api/src/chain.ts
```diff
@@ -42,66 +47,143 @@ export function getAccount(privateKey?: string) {
...
-    const pipeName =
-      args.event?.name || args.address?.substring(0, 8) || "unknown";
-
-    // TODO: save in DB, dont load from scratch every time
-    // TODO: hack for faster startup
     let fromBlock = 0n;
     if (this.publicClient.chain.id === baseGoerli.id) {
       fromBlock = args.event == null ? 7000000n : 5000000n;
     }
-    fromBlock = 8750000n;
```
nalinbhardwaj, line 70: should we bring this back since we're resetting accounts on next deploy?

PR #234, opened 2023-09-06
android: add onboarding step to resolve key issues

### LINKED ISSUE #215: Android enclave broken on some phone models
## Problem

- Ying Tong: can't send. TODO: get Debug Log
- Lev: can't send, "my fingerprint reader is broken".

~Others people have experienced the `missing getSecurityLevel()` error.~ #193

nalinbhardwaj:
New step flow:

Closes #215 : Both YT's and Lev's issue should be dealt with now.

### DIFF
--- a/apps/daimo-mobile/src/action/useSendAsync.ts
+++ b/apps/daimo-mobile/src/action/useSendAsync.ts
@@ -5,6 +5,7 @@ import { useCallback, useEffect } from "react";
 import { Address } from "viem";
 
 import { ActHandle, SetActStatus, useActStatus } from "./actStatus";
+import { Log } from "../logic/log";
 import { useAccount } from "../model/account";
 import { resync } from "../sync/sync";
 
@@ -45,6 +46,7 @@ export function useSendAsync({
       enclaveKeyName,
       account.address,
       keySlot,
+      account.forceWeakerKeys,
       sendFn
     );
 
@@ -71,12 +73,19 @@ export function useSendAsync({
 export function useWarmCache(
   enclaveKeyName?: string,
   address?: Address,
-  keySlot?: number
+  keySlot?: number,
+  forceWeakerKeys?: boolean
 ) {
   useEffect(() => {
-    if (!enclaveKeyName || !address || !keySlot) return;
-    loadAccount(enclaveKeyName, address, keySlot);
-  }, [enclaveKeyName, address, keySlot]);
+    if (
+      !enclaveKeyName ||
+      !address ||
+      !keySlot ||
+      forceWeakerKeys === undefined
+    )
+      return;
+    loadAccount(enclaveKeyName, address, keySlot, forceWeakerKeys);
+  }, [enclaveKeyName, address, keySlot, forceWeakerKeys]);
 }
 
 const accountCache: Map<[Address, number], Promise<DaimoAccount>> = new Map();
@@ -84,7 +93,8 @@ const accountCache: Map<[Address, number], Promise<DaimoAccount>> = new Map();
 function loadAccount(
   enclaveKeyName: string,
   address: Address,
-  keySlot: number
+  keySlot: number,
+  forceWeakerKeys: boolean
 ) {
   let promise = accountCache.get([address, keySlot]);
   if (promise) return promise;
@@ -95,7 +105,12 @@ function loadAccount(
     );
     const signer: SigningCallback = async (hexTx: string) => {
       return {
-        derSig: await requestEnclaveSignature(enclaveKeyName, hexTx),
+        derSig: await requestEnclaveSignature(
+          enclaveKeyName,
+          hexTx,
+          "Authorize transaction",
+          forceWeakerKeys
+        ),
         keySlot,
       };
     };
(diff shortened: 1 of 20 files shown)


packages/daimo-expo-enclave/src/index.ts
```diff
@@ -39,6 +39,14 @@ export async function getBiometricSecurityLevel(): Promise<BiometricSecurityLeve
   return ExpoEnclaveModule.getBiometricSecurityLevel();
 }
 
+/**
+ * Forcibly use software (non-hardware backed) keys. Used for smooth operation on
+ * Android where unpredictable issues with hardware-backed keys are common.
```
dcposch, line 44: this makes it sound like we always use it on android

packages/daimo-expo-enclave/ios/ExpoEnclaveModule.swift
```diff
@@ -34,6 +34,10 @@ public class ExpoEnclaveModule: Module {
       }
     }
 
+    AsyncFunction("forceFallbackUsage") { () in
+      keyManager = FallbackKeyManager()
```
dcposch, line 38: is it possible to put this into a constructor or `init` so that we initialize `keyManager` once?

having this imperative function feels bad, but OK if it's the only way

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
```diff
@@ -228,6 +260,130 @@ function PageBubble({ count, index }: { count: number; index: number }) {
...
+      }
+    }
+    setLoading(false);
+  };
+
+  return (
+    <View style={styles.onboardingScreen}>
+      <View style={styles.createAccountPage}>
+        <TextH1>
+          <Octicons name="unlock" size={40} color={askWeak ? "red" : "black"} />
```
dcposch, line 318: recommend following instead:

```tsx
<Octicons name={askWeak ? "unlock" : "lock"} size={32} />
```

apps/daimo-mobile/src/action/useSendAsync.ts
```diff
@@ -132,16 +153,27 @@ async function sendAsync(
   }
 }
 
-async function requestEnclaveSignature(enclaveKeyName: string, hexTx: string) {
+export async function requestEnclaveSignature(
+  enclaveKeyName: string,
+  hexTx: string,
+  usageMessage: string,
+  forceWeakerKeys: boolean
```
dcposch, line 160: so i think we now need `forceWeakerKeys` everywhere we use `enclaveKeyName`. i wonder if it's worth extracting an `EnclaveKeyInfo` that looks something like

```
interface EnclaveKeyInfo {
  keyName: string
  keyType: <"software" | "hardware">
}
```

dcposch: bit less code in a lot of places, eg here we'd pass `keyInfo: EnclaveKeyInfo`

since figuring out the `keyType` on existing accounts might be inconvenient in the migration, you could also say

```ts
interface EnclaveKeyInfo {
  keyName: string
  forceWeakerKeys: boolean
}
```

dcposch (commented):
LGTM, left comments

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
```diff
@@ -228,6 +260,130 @@ function PageBubble({ count, index }: { count: number; index: number }) {
...
+          <TextCenter>
+            {!askWeak && (
+              <TextBody>
+                Create a cryptographic key on-device that will authorize your
+                account. This makes your money yours alone.
+              </TextBody>
+            )}
+            {askWeak && (
+              <TextBody>
+                Failed to create a hardware enclave authenticated key. Use
```
dcposch, line 331: might be worth saying something like

> Failed to create hardware authenticated key. This can happen on older Android devices. Use software key?

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
```diff
@@ -228,6 +260,130 @@ function PageBubble({ count, index }: { count: number; index: number }) {
...
+        <TextH1>
+          <Octicons name="unlock" size={40} color={askWeak ? "red" : "black"} />
+        </TextH1>
+        <Spacer h={32} />
+        <View style={ss.container.padH16}>
+          <TextCenter>
+            {!askWeak && (
+              <TextBody>
+                Create a cryptographic key on-device that will authorize your
+                account. This makes your money yours alone.
```
dcposch, line 326: 🚀

packages/daimo-expo-enclave/ios/ExpoEnclaveModule.swift
nalinbhardwaj: Yeah, I also wanted to do that but afaict we have no control over `ExportedModule` life cycles and the constructor/`init` is called before we can inject custom arguments in runtime.

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
nalinbhardwaj: done; changed size to 40 to keep it consistent with "bell" size on notifications screen.

apps/daimo-mobile/src/action/useSendAsync.ts
nalinbhardwaj: done latter since easier + also technically we only need to know whether we need to force the enclave or not. If not we can save a call as it can autodetect key being software or hardware.

apps/daimo-mobile/src/view/screen/OnboardingScreen.tsx
nalinbhardwaj: done

packages/daimo-expo-enclave/src/index.ts
nalinbhardwaj: fixed

PR #243, opened 2023-09-07
test: more contract coverage

dcposch:
### 78%

starting coverage

### contracts to improve

- [x] DaimoAccount
- [x] DaimoAccountFactory
- [x] DaimoEphemeralNote

NameRegistry already has full coverage and P256SHA256 is about to be replaced


dcposch:
### minor bug in `forge coverage`

https://github.com/foundry-rs/foundry/issues/5623#issuecomment-1709718248

dcposch:
80% line, 55% branch (undercount, see above)

image

dcposch:
### 85%+

unfortunately, `forge coverage` appears to be inaccurate for both line and branch coverage

it reports much of `DaimoAccount.sol` as uncovered, including the key `validateUserOp` function. it's easy to verify this is incorrect, though. we call that function explicitly in a test, editing the expected return value causes the test to fail, and another test send simulated userOps thru the entry point contract.

editing `validateUserOp` causes both of those to fail.

tldr; `forge coverage` is bugged

dcposch:
fixes #244

packages/contract/src/DaimoAccount.sol
```diff
@@ -189,37 +232,33 @@ contract Account is BaseAccount, UUPSUpgradeable, Initializable {
...
-     */
+    /// Add a signing key to the account
+    /// @param slot the empty slot index to use for this key
+    /// @param key the P256 public key to add
     function addSigningKey(uint8 slot, bytes32[2] memory key) public onlySelf {
         require(keys[slot][0] == bytes32(0), "key already exists");
         require(key[0] != bytes32(0), "new key cannot be 0");
-        require(numActiveKeys < 255, "too many keys");
-        require(slot < 255, "invalid slot");
+        require(slot < maxKeys, "invalid slot");
```
nalinbhardwaj, line 251: flag maxKey

dcposch:
## 85%+, improved key slot handling, added branch coverage

85% is an undercount, see above

image

PR #253, opened 2023-09-11
feat: upgradeable NameRegistry

dcposch:
- NameRegistry is now Ownable, and accessed thru a TransparentUpgradeableProxy
- owner can `forceRegister`, overriding name squatting.
- TransparentUpgradeableProxy is administered by a ProxyAdmin as recommended by OpenZeppelin
- improved test coverage, including testing upgrade


nalinbhardwaj (approved):

packages/daimo-contract/src/tokenMetadata.ts
```diff
@@ -1,5 +1,4 @@
-// Generated by script/fetchTokenMetadata.ts
-
+// TestUSDC on Base Goerli
```
nalinbhardwaj, line 1: I would keep this comment as before. this file is autogenerated and should not be edited manually (after #255 it also generates paymaster variables)

nalinbhardwaj: going to rebase and revert this fetch script deletion since paymaster stuff also uses that script now (and I dont see any other clean way to get the address).

Issue #228, opened 2023-09-04
AmountInput + Send cleanup


dcposch:
## Goal

- No visible toggle between entering an amount and accepting that amount
  - Currently, shows AmountInput vs TitleAmount, two different font sizes
- Code deduplication and cleanup between Create Request, Create Payment Link, and Send
- Corresponding UX consistency between those screens
- Don't make the user dismiss they keyboard before they can proceed 
  - No more numeric-keyboard `onSubmit` button. Instead, enter amount > tap Send/Create directly

dcposch:
done

PR #260, opened 2023-09-18
Fix cursor off-center

markusbug:
Fixed that the cursor is off-center after a user re-enters something in the field.

### DIFF
--- a/apps/daimo-mobile/src/view/shared/AmountInput.tsx
+++ b/apps/daimo-mobile/src/view/shared/AmountInput.tsx
@@ -78,6 +78,7 @@ export function AmountInput({
       keyboardType="numeric"
       placeholder="0"
       placeholderTextColor={color.gray}
+      multiline
       numberOfLines={1}
       autoFocus={autoFocus == null ? true : autoFocus}
       value={strVal}


apps/daimo-mobile/src/view/shared/AmountInput.tsx
```diff
@@ -78,6 +78,7 @@ export function AmountInput({
       keyboardType="numeric"
       placeholder="0"
       placeholderTextColor={color.gray}
+      multiline
```
dcposch, line 81: hmm so the AmountInput was multiline before, and we removed it due to an issue on iOS 

i'll test it again and add more detail

it's a weird fix in any case, because AmountInput is single line

dcposch:
I'm not sure what you mean by "cursor is off-center after a user re-enters something in the field".

I recommend adding screenshots for this kind of change.

I just tried it

### Before, without `multiline`

Screenshot 2023-09-20 at 21 27 33

### After, with `multiline`

Screenshot 2023-09-20 at 21 27 43

Here, the Done button no longer works.

--


Based on above, I don't think `multiline` is the fix.

You're right, though, that the current amount input looks bad.

We can fix it more comprehensively as described here: #228

PR #267, opened 2023-09-23
contract: use OZ safetransfer instead of token transfer

nalinbhardwaj:
USDC Contract actually appears to hardcode `return true` in both `transfer` and `transferFrom` ([code](https://etherscan.io/address/0xa2327a938febf5fec13bacfb16ae10ecbc4cbdcf#code) around line 820) so this is not a real issue for us, but might as well switch to safe transfer to prevent future footguns.

Veridise Audit ID 6


packages/contract/src/DaimoEphemeralNotes.sol
```diff
@@ -68,6 +71,6 @@ contract EphemeralNotes {
 
         emit NoteRedeemed(note, msg.sender);
         delete notes[_ephemeralOwner];
-        token.transfer(msg.sender, note.amount);
+        token.safeTransfer(msg.sender, note.amount);
     }
```
dcposch, line 75: 👍

dcposch (approved):
LGTM

PR #268, opened 2023-09-23
minor nameregistry logic fixes

nalinbhardwaj:
- prevent null/0 registration (Veridise Audit ID 3)
- make external call regular call (Veridise Audit ID 15)

Didn't deploy since there are other upgradeability fixes to make, we can deploy after those.


dcposch (approved):
LGTM

PR #264, opened 2023-09-22
fix @daimo/userop: remove unnecessary RPC, fix op building

nalinbhardwaj:
tested on simulators txes still work fine and sync etc. looks clean :)

drops dependency on userop.js `Client` object and instead recreates the relevant functionality (`eth_sendUserOperation`) in our own code.

Closes #256

PR #270, opened 2023-09-24
feat: DaimoAccount signature validUntil

dcposch:
## Summary

Veridise Audit Issue 7

Updates the format for Daimo userop signatures, adding a `version` and `validUntil`. This lets us require transfers to settle promptly (or never). This prevents a bundler from potentially delaying an op.

Concretely: it would be a bad user experience if a transfer appeared not to go through, then suddenly went through hours later. After this change, we can have a guaranteed deadline.

## Work

- [x] DaimoAccount refactor, update signature format
- [x] Update AccountSendUserop test
- [x] Add test case exercising validUntil
- [x] Update `daimo-userop` to support deadlines


packages/daimo-userop/src/daimoOpBuilder.ts
```diff
@@ -63,26 +50,45 @@ export class DaimoOpBuilder extends UserOperationBuilder {
...
+
+    const parsedSignature = p256.Signature.fromDER(derSig);
+
+    const hexKeySlot = numberToHex(keySlot, { size: 1 });
+    const hexSig = parsedSignature.toCompactHex();
+
+    ctx.op.signature = [hexVersion, hexValidUntil, hexKeySlot, hexSig].join("");
+  };
+
+  /** Sets user-op payment metadata. This controls how fees are paid. */
```
nalinbhardwaj, line 91: nit: and nonce

packages/daimo-userop/src/daimoOpSender.ts
```diff
@@ -0,0 +1,275 @@
+import { derKeytoContractFriendlyKey } from "@daimo/common";
+import * as Contracts from "@daimo/contract";
+import { Client, ISendUserOperationOpts } from "userop";
+import { UserOperationEventEvent } from "userop/dist/typechain/EntryPoint";
```
nalinbhardwaj, line 4: can this be just `from "userop"`? fine anyway since #264 removes all this code anyway

packages/daimo-userop/src/daimoOpBuilder.ts
dcposch: fixed

packages/daimo-userop/src/daimoOpSender.ts
dcposch: not exported directly. i changed it to `import type` to show its just a type annotation

nalinbhardwaj (approved):

PR #279, opened 2023-09-26
fix: DaimoAccount avoid signature malleability

dcposch:
## Summary

Veridise Audit Issue 21

- **avoid signature malleability**
 DaimoAccount.sol now ensures s <= n/2. Conversely, `daimo-userop` modifies generated signatures, where necessary, to meet that requirement.

- **validateUserOp handle invalid keySlot**
  This wan't one of the audit issues, but @nalinbhardwaj found it while reviewing the above.
  This issue could've broken gas estimation for accounts where keySlot 0 doesn't exist.
  Fix: return the invalid flag, don't revert.
  Also 

- **Enable optimization.** 999,999 runs


packages/contract/src/DaimoAccount.sol
```diff
@@ -178,8 +178,8 @@ contract DaimoAccount is IAccount, UUPSUpgradeable, Initializable, IERC1271 {
         uint256 s = uint256(bytes32(signature[33:65]));
 
         // public key to verify against
+        // if keySlot is invalid, (x,y) will be (0,0) and verification will fail
         uint8 keySlot = uint8(signature[0]);
-        require(keys[keySlot][0] != bytes32(0), "invalid key slot");
```
dcposch, line 182: This `require()` was extraneous and can break fee estimation.

nalinbhardwaj (approved):

PR #282, opened 2023-09-27
feat: scratchpad metrics

dcposch:
```
cd app/scratchpad
npm start metrics
```

### CSV output

```
[METRICS] using Base Goerli
[METRICS] compiling signups 0xF64840839FdcEDcF70f7C65D6cf4E2558d854ba4
[METRICS] compiling USDC transfers 0x1B85deDe8178E18CdE599B4C9d913534553C3dBf

date,signups,transfers
2023-09-10,1,4
2023-09-17,4,37
2023-09-24,9,40

Done
```

### DIFF
--- a/apps/scratchpad/src/index.ts
+++ b/apps/scratchpad/src/index.ts
@@ -1,9 +1,16 @@
 // https://api.pimlico.io/v1/goerli/rpc?apikey=70ecef54-a28e-4e96-b2d3-3ad67fbc1b07
 
 import { AppRouter } from "@daimo/api";
+import { getViemClientFromEnv } from "@daimo/api/src/chain";
+import { CoinIndexer } from "@daimo/api/src/contract/coinIndexer";
+import { NameRegistry } from "@daimo/api/src/contract/nameRegistry";
+import { OpIndexer } from "@daimo/api/src/contract/opIndexer";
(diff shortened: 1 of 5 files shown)


apps/scratchpad/src/index.ts
```diff
@@ -43,11 +51,74 @@ function tsDesc() {
 async function ts() {
   console.log("Hello, world");
 
-  const num =
-    23658957077183673623975163142370986705703709099221578179586031616n;
-  console.log(`Hex length: ` + num.toString(16).length);
-  console.log(`toHex() length: ` + numberToHex(num).length);
-  console.log(`toHex(64) length: ` + numberToHex(num, { size: 64 }).length);
+  const publicClient = createPublicClient({ chain: base, transport: http() });
```
nalinbhardwaj, line 54: dumb q, shouldn't this be `baseGoerli`?

nalinbhardwaj: ah I see, this is only used for the call for the fee estimation question we were wondering about yesterday. recommend putting a comment or moving to its own function for clarity.

nalinbhardwaj (approved):

apps/scratchpad/src/index.ts
dcposch: yes this is just the scratchpad, will rename

PR #284, opened 2023-09-30
fix: check nonceType in DaimoNonceMetadata.fromHex

nalinbhardwaj:
Fixes Veridise Audit ID 23, tested request tracking still work.

### DIFF
--- a/packages/daimo-userop/src/nonce.ts
+++ b/packages/daimo-userop/src/nonce.ts
@@ -29,12 +29,12 @@ export class DaimoNonce {
     return nonce;
   }
 
-  public static fromHex(nonce: Hex): DaimoNonce {
+  public static fromHex(nonce: Hex): DaimoNonce | undefined {
     const paddedNonce = bytesToHex(hexToBytes(nonce), { size: 64 });
     const hexMetadata = paddedNonce.slice(0, 2 + 16) as Hex;
     const metadata = DaimoNonceMetadata.fromHex(hexMetadata);
     const key = `0x${paddedNonce.slice(2 + 16, 2 + 16 + 32)}` as Hex;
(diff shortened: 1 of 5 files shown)


packages/daimo-userop/src/nonce.ts
```diff
@@ -62,9 +62,11 @@ export class DaimoNonceMetadata {
     this.identifier = identifier;
   }
 
-  public static fromHex(hexMetadata: Hex): DaimoNonceMetadata {
+  public static fromHex(hexMetadata: Hex): DaimoNonceMetadata | undefined {
     assert(hexMetadata.length === 16 + 2);
     const nonceType = parseInt(hexMetadata.slice(2, 4), 16);
+    if (!(nonceType in DaimoNonceType) || nonceType === DaimoNonceType.MAX)
```
dcposch, line 68: 👍

dcposch (approved):
LGTM. TBH it's not clear to me that this is necessary--any code that handles DaimoNonceType has to handle `MAX`/`default:` anyway--but i guess it makes the typescript typing more correct.

An alternative would be to relax the typing:

```typescript
export declare class DaimoNonce {
    /** Guaranteed in range [0, 255]. See DaimoNonceType for expected values. */
    metadata: number;
```

... which has the nice property that there's a 1:1 mapping between uint256 nonce and `DaimoNonce`, no undefined

nalinbhardwaj:
> LGTM. TBH it's not clear to me that this is necessary--any code that handles DaimoNonceType has to handle MAX/default: anyway--but i guess it makes the typescript typing more correct.

Fair. Typing as number also will require some Enum conversions at a few places, so I think lets just go with this.

PR #305, opened 2023-10-09
feat: add NFC reader support

dcposch:
## Summary

Reads daimo deep links over NFC. NFC reader mode, so the other side is a device doing HCE (card emulation).

Goal: point-of-sale via `DaimoRequestLink`


dcposch:
## Steps left

- [ ] Fix import issue. see below
- [ ] Test on iOS
- [ ] Test on Android

```
ERROR  Invariant Violation: `new NativeEventEmitter()` requires a non-null argument., js engine: hermes
 ERROR  Invariant Violation: "main" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called., js engine: hermes
```

apps/daimo-mobile/app.json
```diff
@@ -79,7 +79,10 @@
           "faceIDPermission": "Protect Daimo using Face ID."
         }
       ],
-      ["expo-notifications"]
+      [
+        "expo-notifications"
+      ],
+      "react-native-nfc-manager"
```
nalinbhardwaj, line 85: You should add prompt strings in [info.plist](https://docs.expo.dev/versions/latest/config/app/#infoplist) as described in https://github.com/revtel/react-native-nfc-manager#setup

apps/daimo-mobile/src/action/useNFCReader.ts
```diff
@@ -0,0 +1,76 @@
...
+  const nav = useNav();
+  useEffect(() => {
+    console.log("[NFC] init");
+    initNFC(nav);
+  }, []);
+}
+
+async function initNFC(nav: ReturnType<typeof useNav>) {
+  await NfcManager.start();
+  console.log("WTF");
```
nalinbhardwaj, line 20: drop

apps/daimo-mobile/src/action/useNFCReader.ts
```diff
@@ -0,0 +1,76 @@
...
+  const isEnabled = await NfcManager.isEnabled();
+  const isSupported = await NfcManager.isSupported();
+  console.log(`[NFC] isEnabled: ${isEnabled}, isSupported: ${isSupported}`);
+  if (!isEnabled || !isSupported) return;
+
+  setInterval(() => {
+    if (isReadingNfc) return;
+    readNdef(nav);
+  }, 1000);
```
nalinbhardwaj, line 29: can this be made much lower, like 100? atm I think this means that the expected time between being in range of the other device and getting the ding is at least half a second which feels like a lot

nalinbhardwaj (changes requested):

apps/daimo-mobile/src/action/useNFCReader.ts
dcposch: from @markusbug : this is just the reset interval after a successful read

apps/daimo-mobile/app.json
markusbug: Added it in this commit: https://github.com/daimo-eth/daimo/pull/305/commits/47d81118af0069e8a37f72aef7e7621e1b434220

Android doesn't need one, as its not a runtime permission

apps/daimo-mobile/app.json
```diff
@@ -0,0 +1,89 @@
+{
```
nalinbhardwaj, line 1: we moved `app.json` to `app.config.js` in this folder (to support multiple variants for local dev vs. prod app), could we move the delete this and move the `info.plist` (and any other?) change there?

also worth testing a fresh build on EAS when you do / if the permission pops up correctly on the dev build

nalinbhardwaj (approved):
lgtmbc

apps/daimo-mobile/app.json
markusbug: Fixed in https://github.com/daimo-eth/daimo/pull/305/commits/1c863d2e7dc371370f44a249c9c72d2d64deca1e

dcposch:
**Will revisit once we test dedicated NFC hardware.**

PR #314, opened 2023-10-10
add blog nav links to homepage

nalinbhardwaj:


dcposch (approved):
Remove `#team` link so it still fits on small phones. otherwise LGTM

image

PR #318, opened 2023-10-12
never mutate react state, match userops correctly

nalinbhardwaj:
Fixes bug where repeating the same transaction in a few minutes makes it disappear.

PR #321, opened 2023-10-12
Add paymaster fees to ops display

### LINKED ISSUE #298: Show fees on HistoryOpScreen

nalinbhardwaj:
Closes #298 

Based on #318 , merge after that.

### DIFF
--- a/apps/daimo-mobile/src/sync/sync.ts
+++ b/apps/daimo-mobile/src/sync/sync.ts
@@ -143,12 +143,8 @@ async function syncAccount(
   );
 
   // Match pending transfers
-  const nowS = Date.now() / 1e3;
-  const maxPendingAgeS = 5 * 60; // 5 mins
   const stillPending = oldPending.filter(
-    (t) =>
-      syncFindSameOp(t, recentTransfers) == null &&
-      nowS - t.timestamp < maxPendingAgeS
+    (t) => syncFindSameOp(t, recentTransfers) == null
   );
   recentTransfers.push(...stillPending);
 
@@ -252,6 +248,7 @@ function addTransfers(
       blockHash: transfer.blockHash,
       logIndex: transfer.logIndex,
       opHash: transfer.opHash,
+      feeAmount: transfer.feeAmount,
     });
   }
 
--- a/packages/daimo-api/src/contract/coinIndexer.ts
+++ b/packages/daimo-api/src/contract/coinIndexer.ts
@@ -141,6 +141,7 @@ export class CoinIndexer {
 
   /* Attach fee amounts to transfer ops and filter out transfers involving
    * paymaster.
+   * TODO: unit test this function
    */
   private attachFeeAmounts(
     transferOpsIncludingPaymaster: TransferOpEvent[]
@@ -150,15 +151,13 @@ export class CoinIndexer {
     for (const op of transferOpsIncludingPaymaster) {
       if (op.opHash === undefined) continue;
 
-      if (!opHashToFee.has(op.opHash)) {
-        opHashToFee.set(op.opHash, 0);
-      }
+      const prevFee = opHashToFee.get(op.opHash) || 0;
 
       if (op.to === tokenMetadata.paymasterAddress) {
-        opHashToFee.set(op.opHash, opHashToFee.get(op.opHash)! + op.amount);
+        opHashToFee.set(op.opHash, prevFee + op.amount);
       } else if (op.from === tokenMetadata.paymasterAddress) {
         // Account for fee refund
-        opHashToFee.set(op.opHash, opHashToFee.get(op.opHash)! - op.amount);
+        opHashToFee.set(op.opHash, prevFee - op.amount);
       }
     }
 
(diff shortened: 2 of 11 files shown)


apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -143,12 +143,8 @@ async function syncAccount(
   );
 
   // Match pending transfers
-  const nowS = Date.now() / 1e3;
-  const maxPendingAgeS = 5 * 60; // 5 mins
   const stillPending = oldPending.filter(
-    (t) =>
-      syncFindSameOp(t, recentTransfers) == null &&
-      nowS - t.timestamp < maxPendingAgeS
```
nalinbhardwaj, line 151: deleted this conditional, seems like it should never happen (or if it does, its a bug so we should not hide it?)

dcposch: this handles the case where bundler fails to include an op (or it gets reorg-d out or reverts for some reason)

avoids having a op that stays pending indefinitely

find to remove for now -  we can reintroduce this in a more sound way by
- using the `validUntil` on all outgoing ops
- saving that expiration date on the pending op in account storage
- in sync, remove pending ops past their expiration (but only after we've successfully loaded sync data from API--they *can* stay pending indefinitely while you're offline, so they resolve once your back online)

packages/daimo-api/src/contract/coinIndexer.ts
```diff
@@ -82,42 +85,42 @@ export class CoinIndexer {
...
+      );
     }
-    // HACK: Ignore paymaster transfers for now
-    // TODO: Stop doing that and show them in UI
-    ret = ret.filter(
-      (log) =>
-        log.from !== tokenMetadata.paymasterAddress &&
-        log.to !== tokenMetadata.paymasterAddress
+
+    const transferOpsIncludingPaymaster = relevantTransfers.map((log) =>
```
dcposch, line 102: 🚀

packages/daimo-api/src/contract/coinIndexer.ts
```diff
@@ -82,42 +85,42 @@ export class CoinIndexer {
...
-
-    if (
-      blockNumber == null ||
-      blockHash == null ||
-      logIndex == null ||
-      transactionHash == null
-    ) {
-      throw new Error(`pending log ${JSON.stringify(log)}`);
-    }
+    const userOp = this.opIndexer.fetchUserOpLog(transactionHash, logIndex);
```
dcposch, line 117: in-memory join (tx log > op log)

right way to do it for now, but a bit fragile--relies on `opIndexer` always updating first before we get here

that should be fine with the way `pipeLogs` currently works

we can revisit with e2pg in the future > joining happens in db or at the time of block ingestion

packages/daimo-api/src/contract/coinIndexer.ts
```diff
@@ -132,6 +135,48 @@ export class CoinIndexer {
...
       logIndex,
       nonceMetadata,
+      opHash,
     };
   }
+
+  /* Attach fee amounts to transfer ops and filter out transfers involving
+   * paymaster.
+   */
+  private attachFeeAmounts(
```
dcposch, line 145: this would be a nice pure function to unit test

packages/daimo-api/src/contract/coinIndexer.ts
```diff
@@ -132,6 +135,48 @@ export class CoinIndexer {
...
+    transferOpsIncludingPaymaster: TransferOpEvent[]
+  ): TransferOpEvent[] {
+    // Map of opHash to fee amount paid to paymaster address
+    const opHashToFee = new Map<Hex, number>();
+    for (const op of transferOpsIncludingPaymaster) {
+      if (op.opHash === undefined) continue;
+
+      if (!opHashToFee.has(op.opHash)) {
+        opHashToFee.set(op.opHash, 0);
+      }
```
dcposch, line 155:
```
const prevFee = opHashToFee.get(op.opHash) || 0;
```
then you can replace the ` opHashToFee.get(op.opHash)!` exprs below

packages/daimo-api/test/pushNotifier.test.ts
```diff
@@ -66,7 +61,7 @@ test("PushNotifier", async () => {
     assert.strictEqual(output[0].title, "Received $5.00");
     assert.strictEqual(
       output[0].body,
-      "charlie.eth accepted your 5.00 USDC request"
+      "charlie.eth fulfilled your 5.00 USDC request"
```
dcposch, line 64: 👍

dcposch (approved):
looks great

left small suggestions

recommend screenshot for the new op page

apps/daimo-mobile/src/sync/sync.ts
nalinbhardwaj: moved to its own issue #325 but will bring this back in this PR so it’s not weird in the short term

packages/daimo-api/src/contract/coinIndexer.ts
nalinbhardwaj: done

packages/daimo-api/src/contract/coinIndexer.ts
nalinbhardwaj: good point, leaving as a todo for future since there's a good chance this code will go away soon anyway with free paymasters.

nalinbhardwaj:
> recommend screenshot for the new op page

PR #331, opened 2023-10-17
Android: Future proof keys by adding key agreement support

### LINKED ISSUE #329: Add Purpose_AGREE_KEY to android keys for future proofing
https://github.com/daimo-eth/daimo/blob/bce7628cb3879b59e83737b81702c2589b0b182e/packages/daimo-expo-enclave/android/src/main/java/expo/modules/enclave/Android30PlusKeyManager.kt#L48

For when we want to do encrypted memos

nalinbhardwaj:
Closes #329


dcposch (approved):
LGTM

i guess if we ship encrypted memos, we'll have to have a fallback for old android. we can cross that when we get there

PR #332, opened 2023-10-17
web: Fix super.so domain

nalinbhardwaj:
super.so killed our notion bridge because we pointed daimo.xyz to a different DNS sigh

would recommend pushing to prod asap, this is killing our blog and privacy page etc. ATM


dcposch (approved):
LGreatTM

PR #333, opened 2023-10-17
web: disable next js data cache on super.so fetches 

nalinbhardwaj:


dcposch (approved):
Looks fine to me, but I don't think this is it unfortunately...
if it was this, redeploying on render would've fixed it

PR #338, opened 2023-10-20
@daimo/mobile: tab navigation

dcposch:
Screenshot 2023-10-19 at 18 11 45
Screenshot 2023-10-19 at 18 11 42
Screenshot 2023-10-19 at 18 11 47


dcposch:
## Icons

image

apps/daimo-mobile/src/view/TabNav.tsx
```diff
@@ -0,0 +1,186 @@
...
+    tabBarStyle: {
+      height: 72,
+      paddingTop: 16,
+      paddingHorizontal: 16,
+    },
+    tabBarItemStyle: {
+      height: 44,
+    },
+  };
+  const head = { ...all, headerShown: true };
```
nalinbhardwaj, line 65: did you mean to hide the header?

apps/daimo-mobile/src/view/TabNav.tsx
```diff
@@ -0,0 +1,186 @@
...
+  route: RouteProp<ParamListTab, keyof ParamListTab>;
+}): BottomTabNavigationOptions {
+  const all: BottomTabNavigationOptions = {
+    headerShown: false,
+    tabBarStyle: {
+      height: 72,
+      paddingTop: 16,
+      paddingHorizontal: 16,
+    },
+    tabBarItemStyle: {
```
nalinbhardwaj, line 61: nit and maybe just looks weird in the screenshot above but maybe the bottom bar icons/text should be vertically centered?

apps/daimo-mobile/src/view/shared/nav.ts
```diff
@@ -15,18 +15,24 @@ import { Hex } from "viem";
...
+  History: undefined;
+  HistoryOp: { op: OpEvent };
+};
+
+export type ParamListSend = {
   Send: SendNavProp;
-  Withdraw: undefined;
-  Request: undefined;
-  Deposit: undefined;
   Note: { link: DaimoLinkNote };
```
nalinbhardwaj, line 26: should this screen (claim note) live in the receive tab?

packages/daimo-api/src/api/search.ts
```diff
@@ -29,8 +34,9 @@ export async function search(
 
 async function tryGetEnsAddr(prefix: string, vc: ViemClient) {
   if (prefix.length < 3) return null;
+  if (!prefix.includes(".")) return null;
   try {
-    const ensName = normalize(prefix.includes(".") ? prefix : prefix + ".eth");
+    const ensName = normalize(prefix);
```
nalinbhardwaj, line 39: might be misreading this code but I think ATM this would not return the ENS if I type `nibnalin.et` right? only when I do the full `nibnalin.eth`?

nalinbhardwaj (approved):

packages/daimo-api/src/api/search.ts
dcposch: correct

the way it was before was broken:
- typing `ansgar` returns *both* `ansgar` and `ansgar.eth`, which confused some people
- then if you keep typing, `ansgar.` returns nothing

here's how it works now:
- you get typeahead search on Daimo names
- you can enter a full ENS name, and it'll find it
- you can enter a full address, and it'll find it

...if you want, we can typeahead-complete `.eth` as soon as the user types `.`, `.e` or `.et`

(for alt-domains like .cb.id, they'd still have to type the whole thing)

apps/daimo-mobile/src/view/TabNav.tsx
dcposch: yes, some tabs need a header others not (otherwise they get a double, nested header)

apps/daimo-mobile/src/view/TabNav.tsx
dcposch: fixed, updated the screenshot

packages/daimo-api/src/api/search.ts
nalinbhardwaj: sounds good, don’t feel strongly about the type ahead after ., feel free to merge as-is

PR #341, opened 2023-10-23
DaimoPaymaster

nalinbhardwaj:
Tested on testnet, also moves paymaster address from being a environment variable to being server side / synced so we can swap it instantly in future.

TODO:
- [x] deploy mainnet post review


packages/contract/src/DaimoPaymaster.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.12;
+
+import "account-abstraction/core/BasePaymaster.sol";
+
+import "./DaimoAccount.sol";
+
+contract DaimoPaymaster is BasePaymaster {
+    mapping(address => bool) public destWhitelist;
+    mapping(address => bool) public senderBlacklist;
```
dcposch, line 10: I'd put a short comment above saying
- Purpose (free ops on Daimo, especially meta-ops like Add Device)
- In what conditions DaimoPaymaster will sponsor a transaction
- Exceptions when it won't

packages/daimo-api/src/contract/paymaster.ts
```diff
@@ -60,26 +60,25 @@ export class Paymaster {
...
+    // const priceMarkup = await this.vc.publicClient.readContract({
+    //   abi: pimlicoPaymasterAbi,
+    //   address: chainConfig.paymasterAddress,
+    //   functionName: "priceMarkup",
+    // });
+    // const previousPrice = await this.vc.publicClient.readContract({
+    //   abi: pimlicoPaymasterAbi,
+    //   address: chainConfig.paymasterAddress,
+    //   functionName: "previousPrice",
+    // });
```
dcposch, line 73: 👍

packages/contract/src/DaimoPaymaster.sol
```diff
@@ -0,0 +1,76 @@
...
+        for (uint256 i = 0; i < addresses.length; i++) {
+            senderBlacklist[addresses[i]] = isBlacklisted;
+        }
+    }
+
+    function _validatePaymasterUserOp(
+        UserOperation calldata userOp,
+        bytes32 userOpHash,
+        uint256 requiredPreFund
+    ) internal override returns (bytes memory context, uint256 validationData) {
```
dcposch, line 46: so if i understand right, we sponsor ops that:
- call `executeBatch` selector on their own acccount
- each call in the batch is either a self-call or in the address whitelist

...but no way to tell whether the account is a Daimo account (ie, from the DaimoAccountFactory) or not, right?

not necessarily an issue--i dont think anyone will grief us, and our downside is limited if they do.

nalinbhardwaj: i was thinking about adding a check whether an account is from DaimoAccountFactory, but since accounts we make our upgradeable in theory someone could just upgrade them to a different implementation to grief us, so it doesn't add much protection.

packages/contract/src/DaimoPaymaster.sol
nalinbhardwaj: done

PR #364, opened 2023-10-28
Add ability to co-install dev build with prod

nalinbhardwaj:
Tested by installing Daimo Dev app with current Testflight/Play Store and using them simultaneously successfully on iOS and Android.

Have not run/tested a fresh prod build, but maybe we should just do that during release testing since it'll increment versions.


apps/daimo-mobile/eas.json
```diff
@@ -12,7 +12,8 @@
       },
       "env": {
         "DAIMO_DOMAIN": "daimo.xyz",
-        "DAIMO_PASSKEY_DOMAIN": "daimo.xyz"
+        "DAIMO_PASSKEY_DOMAIN": "daimo.xyz",
+        "APP_VARIANT": "sim"
```
dcposch, line 16: nit, `DAIMO_APP_VARIANT` ?

nalinbhardwaj: done

apps/daimo-web/public/.well-known/apple-app-site-association
```diff
@@ -17,12 +18,14 @@
...
-      "JV8PYC9QV4.com.daimo"
+      "JV8PYC9QV4.com.daimo",
+      "JV8PYC9QV4.com.daimo.dev"
     ]
   },
   "webcredentials": {
     "apps": [
-      "JV8PYC9QV4.com.daimo"
+      "JV8PYC9QV4.com.daimo",
+      "JV8PYC9QV4.com.daimo.dev"
```
dcposch, line 28: If I'm understanding right

This means that daimo.xyz links will open in *either* the dev app or the production app

If you have both installed, does it prompt you?

dcposch (approved):
LGTM

apps/daimo-web/public/.well-known/apple-app-site-association
nalinbhardwaj: It picks the one higher up in the list if multiple are installed: https://developer.apple.com/forums/thread/649423

PR #365, opened 2023-10-28
api: health check endpoint

nalinbhardwaj:


packages/daimo-api/src/router.ts
```diff
@@ -51,6 +51,9 @@ export function createRouter(
   const publicProcedure = trpcT.procedure.use(tracerMiddleware);
 
   return trpcT.router({
+    health: publicProcedure.query(async (_opts) => {
+      return "healthy";
```
dcposch, line 55: hahaha

we could just fetch an existing route, but this works too

dcposch (approved):
LGTM

PR #372, opened 2023-10-30
Display first time box correctly on requests

nalinbhardwaj:
Tested in sim:

Screenshot 2023-10-30 at 8 58 32 AM
Screenshot 2023-10-30 at 8 58 12 AM

### DIFF
--- a/apps/daimo-mobile/src/sync/recipients.ts
+++ b/apps/daimo-mobile/src/sync/recipients.ts
@@ -10,6 +10,18 @@ export interface Recipient extends EAccount {
   lastSendTime?: number;
 }
 
+/** Adds lastSendTime to an EAccount using account's recentTransfers */
+export function extendToRecipient(
+  account: Account,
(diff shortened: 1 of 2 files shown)


apps/daimo-mobile/src/sync/recipients.ts
```diff
@@ -10,6 +10,18 @@ export interface Recipient extends EAccount {
   lastSendTime?: number;
 }
 
+/** Adds lastSendTime to an EAccount using account's recentTransfers */
+export function extendToRecipient(
```
dcposch, line 14: i'd call this `addLastSendTime` , current name it's hard to tell what it does

dcposch (approved):

apps/daimo-mobile/src/sync/recipients.ts
nalinbhardwaj: done

PR #389, opened 2023-11-05
Move invite codes to be server-side

nalinbhardwaj:
On local dev, before entering the full string "testnet" it will say "Offline?" since we don't run an API server with chainID 8453 but I think that's fine since it only affects us internally.

Also fix capitalisation bug in key rotation push notifier.

### DIFF
--- a/packages/daimo-api/src/env.ts
+++ b/packages/daimo-api/src/env.ts
@@ -264,5 +264,6 @@ export type ReadOnlyContractType<TAbi extends Abi> = GetContractReturnType<
   PublicClient<Transport, Chain>
 >;
 
-export const DAIMO_INVITE_CODES =
-  process.env.DAIMO_INVITE_CODES || "zuconnect,devconnect";
+export const daimoInviteCodes = new Set(
+  (process.env.DAIMO_INVITE_CODES || "zuconnect,devconnect").split(",")
+);
--- a/packages/daimo-api/src/env.ts
+++ b/packages/daimo-api/src/env.ts
@@ -263,3 +263,6 @@ export type ReadOnlyContractType<TAbi extends Abi> = GetContractReturnType<
   TAbi,
   PublicClient<Transport, Chain>
 >;
+
+export const DAIMO_INVITE_CODES =
+  process.env.DAIMO_INVITE_CODES || "zuconnect,devconnect";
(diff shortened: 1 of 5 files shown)


packages/daimo-api/src/pushNotifier.ts
```diff
@@ -280,7 +280,7 @@ export class PushNotifier {
   getPushMessagesFromKeyRotations(logs: SigningKeyAddedOrRemovedLog[]) {
     const messages: ExpoPushMessage[] = [];
     for (const log of logs) {
-      const addr = log.address;
+      const addr = getAddress(log.address);
```
dcposch, line 283: 👍

packages/daimo-api/src/env.ts
```diff
@@ -263,3 +263,6 @@ export type ReadOnlyContractType<TAbi extends Abi> = GetContractReturnType<
   TAbi,
   PublicClient<Transport, Chain>
 >;
+
+export const DAIMO_INVITE_CODES =
+  process.env.DAIMO_INVITE_CODES || "zuconnect,devconnect";
```
dcposch, line 268: nit but i'd do something like

```ts
export const daimoInviteCodes = new Set(
  (process.env.DAIMO_INVITE_CODES || "zuconnect,devconnect").split(',')
);
```

packages/daimo-api/test/pushNotifier.test.ts
```diff
@@ -14,9 +14,9 @@ import { NoteOpLog } from "../src/contract/noteIndexer";
 import { OpIndexer } from "../src/contract/opIndexer";
 import { PushNotifier } from "../src/pushNotifier";
 
-const addrAlice = "0x061b0a794945fe0Ff4b764bfB926317f3cFc8b94";
-const addrBob = "0x061b0a794945fe0Ff4b764bfB926317f3cFc8b93";
-const addrCharlie = "0x061b0a794945fe0Ff4b764bfB926317f3cFc8b95";
+const addrAlice = getAddress("0x061b0a794945fe0Ff4b764bfB926317f3cFc8b94");
+const addrBob = getAddress("0x061b0a794945fe0Ff4b764bfB926317f3cFc8b93");
+const addrCharlie = getAddress("0x061b0a794945fe0Ff4b764bfB926317f3cFc8b95");
```
dcposch, line 19: 👍

dcposch (approved):
LGTM

packages/daimo-api/src/env.ts
nalinbhardwaj: done

PR #407, opened 2023-11-09
Fix history op wrongly syncing

nalinbhardwaj:
Tested on sim.

In future we should take a pass at typing OpEvents better to prevent these.

This PR assumes no transfer page needs to be updated / can never be pending if it does not have an OpHash.

### DIFF
--- a/apps/daimo-mobile/src/sync/sync.ts
+++ b/apps/daimo-mobile/src/sync/sync.ts
@@ -180,7 +180,7 @@ function applySync(account: Account, result: AccountHistoryResult): Account {
   // TODO: store validUntil directly on the op
   const stillPending = oldPending.filter(
     (t) =>
-      syncFindSameOp(t.opHash!, recentTransfers) == null &&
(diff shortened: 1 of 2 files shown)


apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -237,10 +238,10 @@ function applySync(account: Account, result: AccountHistoryResult): Account {
 }
 
 export function syncFindSameOp(
-  op: TransferOpEvent,
+  opHash: Hex,
```
dcposch, line 241: nit but i'd do
`opHash?: Hex` here

then
```
if (opHash == null) return null
```

then you dont need the `!` above

& HistoryOpScreen can be just

```ts
op = syncFindSameOp(op.opHash, account.recentTransfers) || op;
```

dcposch (approved):
LGreatTM

apps/daimo-mobile/src/sync/sync.ts
nalinbhardwaj: done

PR #408, opened 2023-11-09
mobile: fix Android offline state, add fetch timeout

dcposch:
# Before

Android "offline" label hidden behind the camera on Pixel 2. Camera apparently not included in `useSafeAreaInsets`

Android offline unreliable--`fetch()` default timeout is super long, 5 minutes in Chrome, so sync can take forever to fail

# After

- Added manual top padding for Android
- Added fetch timeout

## Offline

Screenshot 2023-11-09 at 12 45 31

## Online

Screenshot 2023-11-09 at 12 48 09

### DIFF
--- a/apps/daimo-mobile/src/logic/trpc.tsx
+++ b/apps/daimo-mobile/src/logic/trpc.tsx
@@ -48,8 +48,8 @@ function getOpts(daimoChain: DaimoChain) {
           daimoChain === "base"
             ? apiUrlMainnetWithChain
             : apiUrlTestnetWithChain,
-        fetch: (input: RequestInfo | URL, init?: RequestInit) => {
-          console.log(`[APP] trpc fetching ${input}`, init);
+        fetch: async (input: RequestInfo | URL, init?: RequestInit) => {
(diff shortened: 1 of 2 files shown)


nalinbhardwaj (approved):

apps/daimo-mobile/src/logic/trpc.tsx
```diff
@@ -59,11 +59,23 @@ function getOpts(daimoChain: DaimoChain) {
           init.headers["x-daimo-platform"] = platform;
           init.headers["x-daimo-version"] = version;
 
-          return fetch(input, init).then((res) => {
+          // Fetch timeout: 5 seconds
+          const timeout = 5000;
```
nalinbhardwaj, line 63: might be too low for some edge cases: https://ui.honeycomb.io/daimo/environments/prod/datasets/daimo-api/result/qyMZ51U5haA?tab=events has P999 of 10-12k ms in a few spikes

dcposch: increased to 10s

outliers should be fixed by retrying

Issue #424, opened 2023-11-13
Incident: timeouts when creating or joining an account


dcposch:
## Summary

<img src="https://github.com/daimo-eth/daimo/assets/169280/beca2f2c-c00a-49b9-8109-45a732cdfc16" width="320" />

> Stuck on this screen

## Timeline

Approximate Turkey time
- 2023-11-13 0100 - Ansgar sends screenshot above
- 2023-11-13 0200 - Filed issue #421
- 2023-11-13 0300 - API deployment https://github.com/daimo-eth/daimo/pull/422
- 2023-11-13 1400 - issue reproduced again, both for account creation and joining an existing account
- 2023-11-13 1500 - honeycomb investigation, see below
- 2023-11-13 1530 - discovered cause for `getAccountHistory`: excessive parallel uncached ENS lookups + Tenderly can't handle it.
- 2023-11-13 1600 - discovered cause for `deployWallet`: `waitForTransactionReceipt` is hanging indefinitely / flaking. not yet certain why, but likely related to the switch from Quiknode to Tenderly. 
- 2023-11-13 1500 - `daimo-api-prod` service restarted, OOM.  unrelated to above, but another source of API unreliability. upgraded instance.

## Investigation

- Cause 1: adding a 10s timeout to all queries
- Cause 2: certain API calls, including `deployWallet` (account create) and `getAccountHistory` with sinceBlockNum 0 (account join) have been getting slower / flaking altogether.

See the new  **Slow RPC** query on our [honeycomb board](https://ui.honeycomb.io/daimo/environments/prod/board/HyhKWdBSyfU).

Example failing account join. 10s client timeout, 11s `getAccountHistory`:
- https://ui.honeycomb.io/daimo/environments/prod/datasets/daimo-api/result/C5gM7FAYfg/trace/mjM89KJof4r?fields[]=s_name&fields[]=s_serviceName&span=0ba10fc05c56ffc9

## Fix

- [x] **Hotfix server-side performance.**
   - [x] Test switching back from Tenderly to Quiknode. Faster `deployWallet`?
      -  **Much faster.** `deployWallet` ~6s to sub 2s
- [x] **Future-proof API performance.**
   - [x] Optimize `deployWallet`. Remove waitForReceipt after adding parallelisation to ensure no nonce conflicts. Verify that we can deploy multiple accounts in a single block.
   - [x] Optimize `getAccountHistory`. Fix excessive ENS lookups issue.
[...]

PR #425, opened 2023-11-13
api performance upgrades

dcposch:
## api: perf

see #424 

- add ENS caching > support accounts with long history
- fix failed `deployWallet` under concurrency > support higher-thruput signups

## web: better copy on the bridge page

d1


packages/daimo-api/src/network/viemClient.ts
```diff
@@ -231,6 +239,60 @@ export class ViemClient {
...
+        return await this.l1Client.getEnsAddress(a);
+      } catch (e: any) {
+        console.log(`[CHAIN] getEnsAddr ${a.name} error: ${e.message}`);
+        return null;
+      }
+    },
+    ({ name }: { name: string }) => name
+  );
+
+  getEnsName = memoize((a: { address: Address }) =>
```
nalinbhardwaj, line 255: add a key function the same as above memoize

nalinbhardwaj (approved):

PR #428, opened 2023-11-14
mobile: no autofocus when tabbing

dcposch:
https://github.com/daimo-eth/daimo/assets/169280/801ff89c-3bc2-4320-b2b2-b48567e90ca0


apps/daimo-mobile/src/view/screen/receive/ReceiveScreen.tsx
```diff
@@ -82,25 +92,28 @@ function RequestScreenInner({ account }: { account: Account }) {
     <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
       <View style={ss.container.screen}>
         <ScreenHeader title="Request" onExit={useExitToHome()} />
-        <Spacer h={64} />
+        <Spacer h={96} />
```
wojtus7, line 95: I think we will need to change those high value paddings to something more based on screen size because if it fix how things look on big phones eg. iPhone 15 Pro things may look wacky on smaller phones like iPhone SE

But probably thing for different issue - not a blocker here 🫥

wojtus7 (approved):
✨

PR #435, opened 2023-11-17
Flatlist inside Bottom Sheet 

wojtus7:
Use flatlist instead of scrollview inside bottom sheet on home screen.

Pls test with wallet with a lot of transactions 🙏


apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
```diff
@@ -127,7 +116,7 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
         >
           {itemMini}
         </Animated.View>
-        {isMini ? itemFullPreview : itemFull}
+        {itemFull}
```
dcposch, line 119: 🚀

apps/daimo-mobile/src/view/shared/HistoryList.tsx
```diff
@@ -80,17 +90,37 @@ export function HistoryListSwipe({
...
+        isHeader: true,
+        month,
+        id: month,
+      });
       lastMonth = month;
     }
-    rows.push(renderRow(t));
+    rows.push({
+      isHeader: false,
+      id: `${t.timestamp}_${t.amount}_${t.blockHash}`,
```
dcposch, line 102: recommend using `${t.timestamp}-${t.from}-${t.to}-${t.txHash}-${t.opHash}` as in renderRow , otherwise we'll have edge cases when multiple txs with same amount end up in same block

also you can deduplicate eg with function `getKey(t: Transfer)`

dcposch (approved):
LGTM after comment

PR #436, opened 2023-11-17
api: add name blacklist

dcposch:
prevent impersonation

onchain name registry remains permissionless and is unaffected by this change.

this is purely removing a list of names (like fake `vitalik`) from search results and push notifications


dcposch:
## Before
`vitalik` shows in search
can send request links

## After
no longer in search
notifications still send, but no longer show as "from vitalik"
request links still work & go to the right place, but no longer show name

image image

nalinbhardwaj (approved):
this change is fine but it won’t remove it for people already sent to whose devices would have cached the name registry?

maybe we should add some regular refresh mechanism to that cache, can be left for seperate PR

PR #443, opened 2023-11-19
Use Pimlico RPC for fee estimation

nalinbhardwaj:
doesn't solve low gas issue yet (https://basescan.org/tx/0x476163c19a5456b02034fd2b8f4488256c66d48c18f9bc9657e76049be286ead)

### DIFF
--- a/packages/daimo-api/src/network/bundlerClient.ts
+++ b/packages/daimo-api/src/network/bundlerClient.ts
@@ -6,6 +6,17 @@ interface GasEstimate {
   preVerificationGas: Hex;
 }
 
+interface GasPriceParams {
+  maxFeePerGas: Hex;
+  maxPriorityFeePerGas: Hex;
+}
+
+interface GasPrice {
+  slow: GasPriceParams;
+  standard: GasPriceParams;
+  fast: GasPriceParams;
+}
+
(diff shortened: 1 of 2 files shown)


packages/daimo-api/src/network/bundlerClient.ts
```diff
@@ -37,6 +48,17 @@ export class BundlerClient {
...
+
+  async getUserOperationGasPriceParams() {
+    const gasPrice = (await this.provider.send(
+      "pimlico_getUserOperationGasPrice",
+      []
+    )) as GasPrice;
+    console.log(
+      `[BUNDLER] fetched gas price params: ${JSON.stringify(gasPrice)}`
+    );
+    return gasPrice.fast;
```
dcposch, line 60: 👍

dcposch (approved):
LGTM

seems like the root cause is pimlico returning too low preVerificationGas?

nalinbhardwaj:
> seems like the root cause is pimlico returning too low preVerificationGas?

preVerificationGas looks correctly set to me. maxFeePerGas and priority fee are lower by ~10x than what the bundler actually sets afaict? (and that diff remains true even after this change)

dcposch:
cherry-picked into #451

PR #447, opened 2023-11-20
No more flatlist inside scrollview

wojtus7:
No more warnings on home screen because no more flatlist inside scrollview.

I use reanimated to achieve exactly same result, but I don't know if I like the code more. No more angry warnings though

### DIFF
--- a/apps/daimo-mobile/src/view/screen/HomeScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/HomeScreen.tsx
@@ -1,16 +1,21 @@
 import Octicons from "@expo/vector-icons/Octicons";
+import { SCREEN_WIDTH } from "@gorhom/bottom-sheet";
 import { useIsFocused } from "@react-navigation/native";
 import { useCallback, useEffect, useRef, useState } from "react";
 import {
   Dimensions,
   Platform,
   RefreshControl,
-  ScrollView,
   StyleSheet,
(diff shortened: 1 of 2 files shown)


apps/daimo-mobile/src/view/screen/HomeScreen.tsx
```diff
@@ -125,13 +149,12 @@ function HomeScreenInner({ account }: { account: Account }) {
...
-          height:
-            screenDimensions.height -
-            tabBarHeight -
-            (Platform.OS === "android" ? 16 : 0),
+          height: screenHeight,
         }}
         onScrollBeginDrag={onScrollBeginDrag}
         onScrollEndDrag={onScrollEndDrag}
+        onScroll={scrollHandler}
+        scrollEventThrottle={8}
```
dcposch, line 157: any reason to make this lower than 16?

wojtus7: Talked Reanimated maintainer asking what will happen on 120hz screens if it only happened every 16ms, and the answer was: choppy experience

In the new release of reanimated it gonna adapt to the screen refresh rate but for 3.3.0 the best idea is to use 8 for now

dcposch: Sounds good!

PR #454, opened 2023-11-22
Remove additional padding when offline

wojtus7:
Add flag to remove additional padding when offline banner appear


dcposch (approved):
LGTM

PR #467, opened 2023-11-28
settings: disable tab swipe inside stack navigator screens

nalinbhardwaj:
Checked on iOS sim


nalinbhardwaj:
added another commit adding back pager view and tab view: they seem like unused dependencies in build but @wojtus7 noted they were mentioned as required dependencies in the docs so we'll keep them anyway https://reactnavigation.org/docs/material-top-tab-navigator

apps/daimo-mobile/package.json
```diff
@@ -76,11 +76,13 @@
     "react-native-gesture-handler": "~2.12.0",
     "react-native-get-random-values": "~1.9.0",
     "react-native-mmkv": "^2.8.0",
+    "react-native-pager-view": "^6.2.2",
     "react-native-qrcode-svg": "^6.2.0",
     "react-native-reanimated": "~3.3.0",
     "react-native-safe-area-context": "4.6.3",
     "react-native-screens": "^3.27.0",
     "react-native-svg": "13.9.0",
+    "react-native-tab-view": "^3.5.2",
```
dcposch, line 85: Do we need these two? they look unused

dcposch (approved):
Otherwise LGTM

PR #484, opened 2023-11-29
mobile: useSendDebugLog

dcposch:


apps/daimo-mobile/src/debugLog.ts
```diff
@@ -38,3 +50,77 @@ export function getDebugLog(headerLines: string[]) {
...
+
+  useEffect(() => {
+    getHardwareSec().then(setSec);
+  }, []);
+
+  if (account == null) {
+    account = getAccountManager().currentAccount;
+  }
+  if (account == null) {
+    return [async () => {}, {}];
```
nalinbhardwaj, line 71: might be misunderstanding something but wouldn't this mean it doesn't work during onboarding? would be nice if it did

nalinbhardwaj (approved):
lgtmbc

PR #485, opened 2023-11-30
daimo-api: use a shovel database to load decoded logs

ryandotsmith:
I've been playing around with data flow today. I like this approach so far. Here's what I like about it:

1. Locating the SQL for retrieving transfers in the same class/file as the CoinIndexer
2. Having the TransferLog type in the same file as the function that constructs TransferLog[]
3. The Watcher (albeit incomplete) is only concerned with tracking the difference between memory and shove's database.

I'm inclined to remove the *Log types (where possible) and instead introduce a new, smaller type. The motivation for this would be to remove extra log fields. For example, s/TransferLog/Transfer. Transfer would still have things like tx_hash, block_hash, etc. but it would not have log.data or log.topics. It didn't make sense to me to carry this data around since it is already decoded in the form of transfer.{from,to,value}.


ryandotsmith:
I went ahead and tried out s/TransferLog/Transfer. I am liking it. Curious what @dc and @nalinbhardwaj think.

packages/daimo-api/src/shovel/watcher.ts
```diff
@@ -0,0 +1,23 @@
+import { Pool } from "pg";
+
+type callback = (pg: Pool, from: bigint, to: bigint) => void | Promise<void>;
+
+export class Watcher {
```
dcposch, line 5: i'd add a comment describing how Watcher works

what happens when a new block appears, etc

packages/daimo-api/src/server/server.ts
```diff
@@ -44,6 +45,10 @@ async function main() {
   const accountFactory = new AccountFactory(vc);
   const crontab = new Crontab(vc, coinIndexer, nameReg, monitor);
 
+  const shovelWatcher = new Watcher();
+  shovelWatcher.addCallback(coinIndexer.load);
+  setInterval(() => shovelWatcher.run(), 1000);
```
dcposch, line 50: nice
so watcher polls, and then:
- poll latest block from pg where we know that all the logs from that block are already ingested
- if > current latest, call all callbacks passing (current, latest)
- each of those goes to an indexer, which loads the logs it needs in that range & processes them
- since we register the indexers in a specific order, the ones that depend on other indexers stay happy

dcposch: once we have it dialed in, poll interval can be much faster

dcposch (commented):
Looks great to me so far!

packages/daimo-api/src/shovel/watcher.ts
ryandotsmith: 100%. This is very much a sketch. Once things start to solidify I will add some docs.

packages/daimo-api/src/contract/noteIndexer.ts
```diff
@@ -1,153 +1,136 @@
...
+        encode(redeemer, 'hex') as "redeemer",
+        encode(ephemeral_owner, 'hex') as "ephemeral_owner",
+        amount,
+      from note_redeemed
+      where block_num >= $1 and block_num <= $2
+    `,
+      [from, to]
+    );
+    const logs = result.rows.map(async (row) => {
+      console.log(`[NOTE] NoteRedeemed ${row.ephemeral_owner}`);
```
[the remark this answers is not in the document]
ryandotsmith: I think this is achieved in the calling function: https://github.com/daimo-eth/daimo/blob/048b95521644e43e3b8fbac89e5dd4b5bb9e5137/packages/daimo-api/src/contract/noteIndexer.ts#L29-L37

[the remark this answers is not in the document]
ryandotsmith: I assume the dependency here is the coordination around `private notes: Map<Address, DaimoNoteStatus> = new Map();`. Since loadCreated is setting data on this map prior to loadRedeemed getting data I think we are good.

packages/daimo-api/src/contract/coinIndexer.ts
```diff
@@ -5,45 +5,71 @@ import {
...
+          log_idx,
+          encode(log_addr, 'hex') as log_addr,
+
+          encode(f, 'hex') as "from",
+          encode(t, 'hex') as "to",
+          v as "value"
+        from transfers
+        where block_num >= $1
+        and block_num <= $2
+        and log_addr = '\\x833589fcd6edb6e08f4c7c32d4f71b54bda02913';
```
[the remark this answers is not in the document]
ryandotsmith: this has been done

PR #501, opened 2023-12-05
Keep splash screen a little longer waiting for sync

wojtus7:
Keep splash screen a little longer waiting for sync or offline banner

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/0ff5a41d-e951-453d-9a40-c933c9755897

Android:

https://github.com/daimo-eth/daimo/assets/42337257/75fe22a6-a5f3-41b1-9fe3-3325768b198d

### DIFF
--- a/apps/daimo-mobile/src/sync/sync.ts
+++ b/apps/daimo-mobile/src/sync/sync.ts
@@ -26,13 +26,16 @@ import { Account, getAccountManager } from "../model/account";
 export function startSync() {
   console.log("[SYNC] APP LOAD, starting sync");
   maybeSync(true).then((success) => {
-    if (success) {
-      // create small delay to let interface render becuase this callback
-      // run right after getting data so the interface still got to adapt
-      setTimeout(() => {
-        SplashScreen.hideAsync();
-      }, 300);
+    if (!success) {
+      updateNetworkState(() => {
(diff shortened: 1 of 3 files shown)


apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -24,7 +25,15 @@ import { Account, getAccountManager } from "../model/account";
...
-  maybeSync(true);
+  maybeSync(true).then((success) => {
+    if (success) {
+      // create small delay to let interface render becuase this callback
+      // run right after getting data so the interface still got to adapt
+      setTimeout(() => {
+        SplashScreen.hideAsync();
+      }, 300);
+    }
+  });
```
dcposch, line 36: hmm we want to hide the splash screen as soon as possible

we could do it in `finally` here regardless of whether that initial sync succeeded

(we can update `useNetworkState()` so that the app shows offline as soon as initial sync fails. either here, or i'm happy to do it in a future PR.)

additionally: in `App` render > useEffect, hide the splashscreen as soon `account.lastBlockTimestamp` is less than 10 minutes ago. this means that it'll hide instantly once we're rendering with fresh data (and the setTimeout one here is just a failsafe)

dcposch (approved):
LGTM after comment

once done no need for re-review, ship it 🚢

PR #506, opened 2023-12-08
api: use bulk, send compressed bundles

dcposch:
## Goal

Cheap 4337 on L2.

## How it works

See https://github.com/daimo-eth/bulk

In particular, see `BundleBulker.t.sol` for before-and-after compression details.

## Results

Submitting a 4337 bundle containing a single transfer, with passkey signature.

**Before: [1540 bytes calldata, 15k L1 gas](https://goerli.basescan.org/tx/0x6816b60541a802e346071392f60c8c79b178dc346f366f3f2e5849a024c14fe5)**

Transaction is submitted directly to the EntryPoint.

**After: [356 bytes calldata, 8k L1 gas](https://goerli.basescan.org/tx/0x3320191b32218c884ab22ec278cad9a0da18de50424449e0cad33c092927682b)**

Transaction is submitted via [bulk](https://github.com/daimo-eth/bulk), which decompresses the bundle and then calls the EntryPoint.

Additional optimization possible.

We'll also see automatic additional gains with scale. Once we're averaging multiple ops per second, we'll have more than one op per bundle, amortizing bundle overhead.

## Follow-up

**Further optimization possible in a future PR:**

- [ ] **Remove paymaster signature.** Sponsoring paymasters currently sign over each op in order to avoid being griefed, but this adds an additional 64+ bytes calldata to each op. We can solve this with a stateful paymaster.
- [ ] **Indexed accounts.** We can save an additional ~32 bytes by going from a 20-byte from and to address to a 4-byte index, for the common case of a transfer between two daimo accounts.

A few more optimizations possible after that, including tick-based gas pricing (1 byte per gas constant rather than 4-6) but we approach diminishing returns.

We can reprioritize small optimizations if we hit enough scale to justify signature aggregation, at which point they become significant again.


packages/daimo-api/src/network/bundlerClient.ts
```diff
@@ -21,12 +43,183 @@ interface GasPrice {
...
+    ret.push(`0x${m[1]}`); // paymaster data
+
+    return concatHex(ret);
+  }
+
+  async sendCompressedOpToBulk(
+    compressed: Hex,
+    walletClient: WalletClient<Transport, Chain, Account>
+  ) {
+    const txHash = await walletClient.writeContract({
```
nalinbhardwaj, line 213: shouldn't this go through the `ViemClient` instead (to prevent nonce reuse as in deploys?)

packages/daimo-api/src/network/bundlerClient.ts
```diff
@@ -21,12 +43,183 @@ interface GasPrice {
...
+    ret.push(`0x${m[2]}`); // sig r
+    ret.push(`0x${m[3]}`); // sig s
+    ret.push(`0x${m[4]}`); // authenticator challenge
+
+    // paymaster signature, if present
+    const paymasterAddr = this.inflatorPaymaster;
+    const paymasterRegex = new RegExp(`${paymasterAddr}(.*)$`, "i");
+    m = paymasterRegex.exec(op.paymasterAndData);
+    if (!m) throw new Error("can't compress, bad paym.:" + op.paymasterAndData);
+    ret.push(`0x${m[1]}`); // paymaster data
```
nalinbhardwaj, line 204: very dense logic, maybe add a dummy test just to prevent any obvious breaks in future?

dcposch:
## Prod

50% cheaper.

**Before: [11,964 L1 gas, $0.76](https://basescan.org/tx/0x74db58fa320a5a710bd0f4d0a709d33dad2644a1ded8b400748f67def345ccda)**

**After: [6668 L1 gas, $0.38](https://basescan.org/tx/0x5858378befbba1d8ac7952415d73b16808fec5e322511ad0fab7654b01017622)**

PR #511, opened 2023-12-09
AccountScreen

dcposch:
image image image image

https://github.com/daimo-eth/daimo/assets/169280/5234db43-401c-45f1-9a95-5250fc34ad9f

### DIFF
--- a/apps/daimo-mobile/src/view/TabNav.tsx
+++ b/apps/daimo-mobile/src/view/TabNav.tsx
@@ -146,6 +146,7 @@ function SendTab() {
         <SendStack.Screen name="QR" component={QRScreen} />
         <SendStack.Screen name="SendLink" component={SendNoteScreen} />
         <HomeStack.Screen name="Account" component={AccountScreen} />
+        <HomeStack.Screen name="HistoryOp" component={HistoryOpScreen} />
       </SendStack.Group>
     </SendStack.Navigator>
   );
@@ -159,8 +160,8 @@ function HomeTab() {
       <HomeStack.Group>
         <HomeStack.Screen name="Home" component={HomeScreen} />
         <HomeStack.Screen name="QR" component={QRScreen} />
-        <HomeStack.Screen name="HistoryOp" component={HistoryOpScreen} />
         <HomeStack.Screen name="Account" component={AccountScreen} />
+        <HomeStack.Screen name="HistoryOp" component={HistoryOpScreen} />
       </HomeStack.Group>
     </HomeStack.Navigator>
   );
--- a/apps/daimo-mobile/src/view/TabNav.tsx
+++ b/apps/daimo-mobile/src/view/TabNav.tsx
@@ -13,6 +13,7 @@ import { useEffect, useState } from "react";
 import { Platform } from "react-native";
 import { EdgeInsets, useSafeAreaInsets } from "react-native-safe-area-context";
 
+import { AccountScreen } from "./screen/AccountScreen";
 import { AddDeviceScreen } from "./screen/AddDeviceScreen";
 import { AddPasskeyScreen } from "./screen/AddPasskeyScreen";
 import { DeviceScreen } from "./screen/DeviceScreen";
@@ -158,6 +159,7 @@ function HomeTab() {
         <HomeStack.Screen name="Home" component={HomeScreen} />
         <HomeStack.Screen name="QR" component={QRScreen} />
         <HomeStack.Screen name="HistoryOp" component={HistoryOpScreen} />
+        <HomeStack.Screen name="Account" component={AccountScreen} />
       </HomeStack.Group>
     </HomeStack.Navigator>
   );
--- a/apps/daimo-mobile/src/view/TabNav.tsx
+++ b/apps/daimo-mobile/src/view/TabNav.tsx
@@ -145,6 +145,7 @@ function SendTab() {
         <SendStack.Screen name="SendTransfer" component={SendTransferScreen} />
         <SendStack.Screen name="QR" component={QRScreen} />
         <SendStack.Screen name="SendLink" component={SendNoteScreen} />
+        <HomeStack.Screen name="Account" component={AccountScreen} />
       </SendStack.Group>
     </SendStack.Navigator>
   );
--- a/apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
@@ -258,7 +258,7 @@ function RecipientDisplay({
 
   const nav = useNav();
   const goToAccount = useCallback(() => {
-    nav.navigate("HomeTab", {
+    nav.navigate("SendTab", {
       screen: "Account",
       params: { eAcc: recipient },
     });
--- a/apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
@@ -22,6 +22,7 @@ import { Recipient, addLastSendTime } from "../../../sync/recipients";
 import { AccountBubble } from "../../shared/AccountBubble";
 import { AmountChooser } from "../../shared/AmountInput";
 import { ButtonBig } from "../../shared/Button";
+import { ButtonCircle } from "../../shared/ButtonCircle";
 import { InfoBox } from "../../shared/InfoBox";
 import { ScreenHeader } from "../../shared/ScreenHeader";
 import Spacer from "../../shared/Spacer";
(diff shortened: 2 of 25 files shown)


apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
```diff
@@ -255,9 +256,19 @@ function RecipientDisplay({
   const { originalMatch } = recipient;
   const showOrig = originalMatch != null && originalMatch !== disp;
 
+  const nav = useNav();
+  const goToAccount = useCallback(() => {
+    nav.navigate("HomeTab", {
```
nalinbhardwaj, line 261: This transition from send tab -> home tab feels a bit weird in the video, thoughts on duping the screen in send tab? cc @michelle-chiu as well

apps/daimo-mobile/src/view/shared/useSwipeUpDown.tsx
```diff
@@ -0,0 +1,95 @@
...
+    (Platform.OS === "android" ? 16 : 0);
+
+  // Hide bottom sheet when tapping a bottom tab.
+  const nav = useNav();
+  const bottomSheetRef = useRef<SwipeUpDownRef>(null);
+  const isFocused = useIsFocused();
+  useEffect(() => {
+    if (nav.getParent()) {
+      // @ts-ignore
+      const unsub = nav.getParent().addListener("tabPress", (e: Event) => {
```
nalinbhardwaj, line 40: nit: could remove the `ts-ignore` here by checking / early exiting if `nav.getParent()` is null I think

nalinbhardwaj:
Another idea: we could make nav such that the search at the top on the home tab takes you to the user profile instead of the user send page. Then the user stories are:

- Search on home > User profile with send button
- Click Send on home > search > enter amount to send page

Thoughts?

apps/daimo-mobile/src/view/TabNav.tsx
```diff
@@ -158,6 +159,7 @@ function HomeTab() {
         <HomeStack.Screen name="Home" component={HomeScreen} />
         <HomeStack.Screen name="QR" component={QRScreen} />
         <HomeStack.Screen name="HistoryOp" component={HistoryOpScreen} />
+        <HomeStack.Screen name="Account" component={AccountScreen} />
```
nalinbhardwaj, line 162: also nit but might call this "profile"?

apps/daimo-mobile/src/view/shared/useSwipeUpDown.tsx
dcposch: nah it also doesn't like `tabPress` / thinks that's not an event

apps/daimo-mobile/src/view/TabNav.tsx
dcposch: matched the figma

you can end up on this page for any ethereum account:  daimo, ens, bare address.

dcposch: not yet done: avoid linking to account page for accounts we know you shouldn't send to directly, like `payment link`.

+ relatedly, on a transaction page claiming a payment link, we should link to the original sender of the link.

apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
dcposch: video is edited to fit in the GH 10mb limit

dcposch: i see what you mean though--you're right, i think we can keep the nav smoother that way

dcposch:
> we could make nav such that the search at the top on the home tab takes you to the user profile instead of the user send page

i'll try it and see what it feels like

it's def necessary once we have Request / Send (today, just Send)

apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
dcposch: done

michelle-chiu: @dcposch @nalinbhardwaj - How would we feel about removing the second screen entirely? 

Other than that, looking great so far! Thanks for taking care of this so quickly.

dcposch: what do you mean by the second screen?

michelle-chiu: Also, instead of making "View on Block Explorer" the blue link style, let's keep it consistent with other parts of the app by making it a secondary button style (seen here): 

Screenshot 2023-12-10 at 7 59 02 PM

Figma link: https://www.figma.com/file/wBBNeEzlsObaE8A8i9iVf7/Mobile-App?type=design&node-id=2403%3A14028&mode=design&t=TtmgbNcOvSyhvvXU-1

dcposch:
## Edge case: non-sendable accounts

Users should never send directly to certain accounts, such as the `payment link` or `paymaster` contract.

`HistoryOpScreen` involving these accounts no longer links to the account page.

image

...and if you do somehow get to the account page, eg via a specially crafted deeplink, they no longer have the send button. Before / after:

image image


...and if you somehow get to the send flow, eg via another specially crafted deeplink, you can't send:

image

dcposch:
## New videos

### Home tab to account

https://github.com/daimo-eth/daimo/assets/169280/6032533f-ac9a-4b05-9bc5-8f42d35dbac2

### Send tab to account

https://github.com/daimo-eth/daimo/assets/169280/0ac35898-2da1-443f-af79-a1e1e2a6f558

apps/daimo-mobile/src/view/TabNav.tsx
dcposch: done, see below

apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
dcposch: Sounds good

image

dcposch: i think i see what you mean. you're talking about getting rid of `HistoryOpScreen` (below) and linking directly to `AccountScreen`

two thoughts:

1. HistoryOpScreen could def use a design pass
2. I'm down to bypass it --tap on HomeScreen history, go direct to account.

then if you want the receipt for a particular transaction, you click once on `HomeScreen` > click on that same transaction again on the "Transactions between you" list on the `AccountScreen` > see transaction details

one challenge to solve: if i scroll down on HomeScreen to a transaction that happened a while ago, then I have to repeat that process on AccountScreen to find that same tx again.

LMK if you have ideas.

dcposch:
Hmm, direct-to-account has some issues.

- If I click on a particular transaction, I have to find it again on the "Transfers between us" list
- If I click on a transaction creating a payment link, there is not other account for me to go to (link has not yet been claimed), and going to the `AccountScreen` for the payment link contract address is confusing.

## Direct-to-account

https://github.com/daimo-eth/daimo/assets/169280/92333a87-cee2-4679-b041-b4c82ce49d38

Going to revert for now.

We can improve `HistoryOpScreen` transaction details in next release.

dcposch:
## Copy link to account

https://github.com/daimo-eth/daimo/assets/169280/ced571ae-63fa-429a-a866-1f55545f5eb9

apps/daimo-mobile/src/view/screen/send/SendTransferScreen.tsx
michelle-chiu: @dcposch Yes, exactly—I was implying removing the HistoryOpScreen (above) and linking directly to AccountScreen. 

1. I'm happy to create new designs here for the HistoryOpScreen!
2. Regarding the challenge to solve, do we know if the transaction history is that valuable of a touchpoint to users to have this be a significant issue? I can see the added friction there if users have to find the transaction again, but this may not be as large of an issue we believe it may be.

michelle-chiu:
> Another idea: we could make nav such that the search at the top on the home tab takes you to the user profile instead of the user send page. Then the user stories are:
> 
> * Search on home > User profile with send button
> * Click Send on home > search > enter amount to send page
> 
> Thoughts?



> Another idea: we could make nav such that the search at the top on the home tab takes you to the user profile instead of the user send page. Then the user stories are:
> 
> * Search on home > User profile with send button
> * Click Send on home > search > enter amount to send page
> 
> Thoughts?

Great idea, Nalin! These are the exact user flows that Venmo adopts—I would be happy to move forward with the decisions above. The only drawback I can see is that the profile page may look empty if the two users don't have a decent transaction history between the two of them.

michelle-chiu (approved):

dcposch:
## Added date

Daimo accounts show a "Joined..." date corresponding to the `Registered` event in `NameRegistry`.

All other Ethereum accounts show a summary of the address.

image image

PR #517, opened 2023-12-11
daimo-api: only update latest if progress was made

ryandotsmith:


dcposch (approved):
Add a comment re why index can return a block earlier than the min block passed in

otherwise LGTM

PR #519, opened 2023-12-12
daimo-api: bugfix. return latest block for getAccountHistory

ryandotsmith:
Prior to this commit, and prior to shovel, the getAccountHistory function would use the ViemClient to get the latest block. This assumed that the ViemClient would be updating its block. Now that we are using Watcher, ViemClient was no longer updating its block which means that it latest value was only set on startup and then never updated.

Now you can ask Watcher for the latest block. It makes an RPC request to get the block's timestamp. Ideally this should be in the shovel table but it's not. I'm going to fix shovel so that it stores the block timestamp on the task table so that the Watcher will have the timestamp whenever it has the number.

### DIFF
--- a/packages/daimo-api/src/api/getAccountHistory.ts
+++ b/packages/daimo-api/src/api/getAccountHistory.ts
@@ -15,6 +15,7 @@ import { KeyRegistry } from "../contract/keyRegistry";
 import { NameRegistry } from "../contract/nameRegistry";
 import { Paymaster } from "../contract/paymaster";
 import { ViemClient } from "../network/viemClient";
+import { Watcher } from "../shovel/watcher";
 
 export interface AccountHistoryResult {
   address: Address;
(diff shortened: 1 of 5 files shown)


packages/daimo-api/src/api/getAccountHistory.ts
```diff
@@ -73,10 +75,10 @@ export async function getAccountHistory(
   }
 
   // Get the latest block + current balance.
-  const lastBlk = vc.getLastBlock();
```
nalinbhardwaj, line 76: should we delete this function and associated code in `viemClient.ts`?

ryandotsmith: I've been waiting to delete the unused ViemClient code until I get watcher 100% working

nalinbhardwaj (approved):
lgtm

PR #520, opened 2023-12-12
apps/scratchpad: use timestamp instead of blockNumber

ryandotsmith:


apps/scratchpad/src/index.ts
```diff
@@ -64,10 +64,9 @@ async function metrics() {
   const signups = new Map<string, number>();
   const daimoChain = daimoChainFromId(vc.publicClient.chain.id);
   for (const log of nameReg.logs.sort(
-    (a, b) => Number(a.blockNumber) - Number(b.blockNumber)
+    (a, b) => Number(a.timestamp) - Number(b.timestamp)
```
nalinbhardwaj, line 67: nit: I think you don't need the `Number()` wrappers either now

nalinbhardwaj (approved):
lgtmbc

apps/scratchpad/src/index.ts
ryandotsmith: good catch. ty

PR #521, opened 2023-12-12
daimo-api: wait for deploy to be indexed

ryandotsmith:
I'm 50/50 on this change. I don't like that it adds a delay in the signup process. But it avoids hitting an error (silent?) when the user signs up and their keys have not been indexed.

The alternative is that we remove the assert and instead return some kind of empty result for getAccountHistory until the user's data has been indexed.

https://github.com/daimo-eth/daimo/blob/master/packages/daimo-api/src/api/getAccountHistory.ts#L107-L108

### DIFF
--- a/packages/daimo-api/src/shovel/watcher.ts
+++ b/packages/daimo-api/src/shovel/watcher.ts
@@ -44,10 +44,23 @@ export class Watcher {
   }
 
   async waitFor(blockNumber: bigint, tries: number): Promise<boolean> {
+    const t0 = Date.now();
     for (let i = 0; i < tries; i++) {
-      if (this.latest >= blockNumber) return true;
+      if (this.latest >= blockNumber) {
+        console.log(
+          `[SHOVEL] waiting for block ${blockNumber}, found after ${
+            Date.now() - t0
+          }ms`
+        );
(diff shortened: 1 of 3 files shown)


packages/daimo-api/src/shovel/watcher.ts
```diff
@@ -43,6 +43,14 @@ export class Watcher {
     };
   }
 
+  async waitFor(blockNumber: bigint, tries: number): Promise<boolean> {
+    for (let i = 0; i < tries; i++) {
+      if (this.latest >= blockNumber) return true;
+      await new Promise((res) => setTimeout(res, 250));
+    }
```
dcposch, line 50: i'd add a log here, eg
```
[SHOVEL] waiting for block 9123456, found after 1250ms
```
or 
```
[SHOVEL] waiting for block 9123456, NOT FOUND, still on 9123455 after 2000ms
```

dcposch (approved):
LGTM after comment

packages/daimo-api/src/shovel/watcher.ts
ryandotsmith: added

PR #533, opened 2023-12-14
api: sort transaction index for key rotations within block

nalinbhardwaj:
Subtle ordering bug if two userops make it into the same block.


dcposch (approved):
LGTM

Issue #513, opened 2023-12-11
Nav: easy way to cancel out of a flow


dcposch:
## Goal

Currently, if I'm somewhere *within* the home tab, then tab over to Send and back to home, I'm back where I left off.

That seems fine... but there's no quick way to get back to the main home screen.

I find myself tapping home several times expecting to get there, and it doesn't work.

## Proposed fix

If I'm already on a tab and I tap that same tab again, it resets that tab to its home state.

CC @michelle-chiu lmk if you have an alternate solution

michelle-chiu:
@dcposch In my original designs, I included a "x" icon on the upper-right hand corner of the screen to allow the user to exit back to the homepage (see screenshot)

<img width="371" alt="Screenshot 2023-12-10 at 10 25 46 PM" src="https://github.com/daimo-eth/daimo/assets/5884103/9b172624-f2f9-475b-a5bb-bfa10ed801ab">


What if we implemented your proposed fix and also the "x?"

dcposch:
done in #536

PR #536, opened 2023-12-14
Add X to go back to home from send flow

wojtus7:
Just used `navResetToHome` func to go back.

Every time we use navResetToHome it blinks a little creating ugly effect that is much more visible on Android. I think we need to find some more elegant way to handle navigation reset. (Will try to investigate that now but since this is not a part of [this issue](https://github.com/daimo-eth/daimo/issues/513), it should be a separate PR)

iOS: 

https://github.com/daimo-eth/daimo/assets/42337257/eb38f0d9-0dcc-4c29-9dc0-e29c9257098e

Android:

https://github.com/daimo-eth/daimo/assets/42337257/30942acd-85ee-460a-b604-5c0ebc434798

### DIFF
--- a/apps/daimo-mobile/src/view/screen/send/SendNoteScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/send/SendNoteScreen.tsx
@@ -40,7 +40,7 @@ export function SendNoteScreen({ route }: Props) {
   return (
     <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
       <View style={ss.container.screen}>
-        <ScreenHeader title="Send Link" onBack={goBack} />
+        <ScreenHeader title="Send Link" onBack={goBack} onExit={goHome} />
         <Spacer h={8} />
         <InfoBox
           title="Pay by sending a link"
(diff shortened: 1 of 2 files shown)


wojtus7:
@nalinbhardwaj Added 🫡

nalinbhardwaj (approved):

apps/daimo-mobile/src/view/screen/send/SendNoteScreen.tsx
```diff
@@ -33,14 +38,18 @@ export function SendNoteScreen({ route }: Props) {
...
+
   const goHome = useExitToHome();
   const goBack = nav.canGoBack() ? nav.goBack : goHome;
   useDisableTabSwipe(nav);
 
   return (
     <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
       <View style={ss.container.screen}>
-        <ScreenHeader title="Send Link" onBack={goBack} />
+        <ScreenHeader title="Send Link" onBack={goBack} onExit={resetHome} />
```
dcposch, line 52: i think we can use `goHome` instead of `resetHome` to avoid the flash?

dcposch (approved):
Updated, simplified. Using `useExitToHome`

(@wojtus7 lmk if there's a reason we have to hard reset with `navResetToHome`)

PR #537, opened 2023-12-14
Change how reset works to avoid blink and reload

wojtus7:
Not a big fan of this change code-wise, but the visual aspect is much better this way.

Also previous version surely will always get the job done even if it looks worse, but with this solution something can go wrong with `popToTop` and `navigate` at the same time 😳

IDK I just hate this reload blink when use `reset`

iOS after:

https://github.com/daimo-eth/daimo/assets/42337257/467c3237-dc2d-4b84-a998-c0f48a80ad63

iOS before:

https://github.com/daimo-eth/daimo/assets/42337257/dc4dc972-02b6-499c-a647-b60d5b32a8ce

Android after:

https://github.com/daimo-eth/daimo/assets/42337257/08fe1f05-822b-47ed-9923-a5aebbb2d8ce

Android before:

https://github.com/daimo-eth/daimo/assets/42337257/b385856d-bc0f-4e00-938e-28a51e64a59f


dcposch (commented):
I wonder if we can get rid of `navResetToHome` entirely and just use `useExitToHome()`, which calls `navigate` instead of `reset`

wojtus7:
> I wonder if we can get rid of `navResetToHome` entirely and just use `useExitToHome()`, which calls `navigate` instead of `reset`

But this way I think we do not reset stack on any open tab so if we are in send flow the send flow won't be reseted - navigate will only take us to home screen and leave send flow wherever it was

dcposch (approved):
Updated to unify `useExitToHome` and `navResetToHome`

both of them now work the new way:
- don't call `nav.reset` to avoid the ugly flash
- do call `nav.popToTop` afterward to avoid dangling state

PR #554, opened 2023-12-19
api: send compressed ops to pimlico

dcposch:
@kristofgaszo


dcposch:
Revisiting once Pimlico has per-op compression

dcposch:
Will revisit once we have signed paymaster tickets.

PR #572, opened 2023-12-22
shovel: separate mainnet/testnet

nalinbhardwaj:
tested with shovel-stage instance, will update environment before pushing to prod

Based on #575 , merge that first.

### DIFF
--- a/packages/daimo-api/shovel/config.json
+++ b/packages/daimo-api/shovel/config.json
@@ -2,28 +2,18 @@
   "pg_url": "$DATABASE_URL",
   "eth_sources": [
     {
-      "name": "base",
-      "chain_id": 8453,
-      "url": "$BASE_RPC_URL",
-      "batch_size": 10,
-      "concurrency": 4
-    },
-    {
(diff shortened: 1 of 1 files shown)


packages/daimo-api/shovel/config.json
```diff
@@ -2,28 +2,18 @@
...
-      "batch_size": 10,
-      "concurrency": 4
-    },
-    {
-      "name": "base-goerli",
-      "chain_id": 84531,
-      "url": "$BASE_GOERLI_RPC_URL",
-      "batch_size": 10,
-      "concurrency": 4
+      "name": "$CHAIN_NAME",
```
dcposch, line 5: could deduplicate this environment variable with `DAIMO_CHAIN` ?

dcposch (approved):

packages/daimo-api/shovel/config.json
nalinbhardwaj: shovel runs in a seperate service so no env vars are shared, will just leave as is

PR #573, opened 2023-12-23
mobile: upgrade HistoryOpScreen

### LINKED ISSUE #553: [Design] Design improvements to transaction details (HistoryOpScreen)
When a user taps on a transaction cell, a bottom sheet will appear with the transaction details. When the contact initial bubble is tapped on, they are sent to the "account page" (see screenshot): 

Screenshot 2023-12-18 at 4 07 44 PM

Figma link: https://www.figma.com/file/wBBNeEzlsObaE8A8i9iVf7/Mobile-App?type=design&node-id=2521%3A16074&mode=design&t=uXCFjvTSz6zvIGbO-1

dcposch:
## New screens

Screenshot 2023-12-23 at 00 35 52 Screenshot 2023-12-23 at 00 36 00

Fixes #553 

## Nav

#553 also specifies nav changes:
- Open in bottom sheet, rather than whole page
- Let user open transaction details directly from homepage, without a nav transition

We can do these as a followup after the holidays. There are some tricky details to making this bottom sheet feel right, especially since it's a *second* bottom sheet with the (potentially expanded) HistoryList underneath.

CC @wojtus7


apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
```diff
@@ -137,123 +147,109 @@ function TransferBody({
...
-  const opRequestId = op.nonceMetadata
-    ? DaimoNonceMetadata.fromHex(op.nonceMetadata)?.identifier.toString()
-    : undefined;
-  const matchingTrackedRequest = account.trackedRequests.find(
-    (req) =>
-      req.requestId === opRequestId &&
-      req.amount === `${op.amount}` &&
-      op.to === account.address
-  );
+  // TODO: show if this transfer filled our request.
```
dcposch, line 150: this is not currently in the designs

leaving the logic here for if we re-add it.

packages/daimo-api/src/server/router.ts
```diff
@@ -56,16 +56,21 @@ export function createRouter(
     return opts.next();
   });
 
+  const readyMiddleware = trpcT.middleware(async (opts) => {
+    if (!notifier.isInitialized) {
+      throw new Error("not ready");
+    }
+    return opts.next();
+  });
```
nalinbhardwaj, line 64: just for my understanding, this is functionally equivalent to previous version right? or is there some edge case this changes?

dcposch: this makes it so *all* routes, not just the health route, return this error before the api server is ready

apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
nalinbhardwaj: doesn't need to be in this pr but imo we can just drop all the current request logic and redo it later when we make requests compatible with non-daimo wallets for the sender.

nalinbhardwaj (approved):
lgtm

packages/daimo-api/src/server/router.ts
nalinbhardwaj: right but no other route should be getting hit before the server is ready and health route is "healthy" right? or did you observe behavior/edge case where it can be?

apps/daimo-mobile/src/view/screen/send/SendNavScreen.tsx
```diff
@@ -20,33 +20,34 @@ type Props = NativeStackScreenProps<ParamListSend, "SendNav">;
 export function SendNavScreen({ route }: Props) {
   const { autoFocus } = route.params || {};
 
+  // Search prefix
+  // Clear prefix on back button
```
dcposch, line 24: snuck in this bugfix:
- click send
- enter a letter, see search results
- click back
- click send again

before: your old search prefix is still there. there is no way to clear out of search without manually backspacing away the query.

new: query is cleared.

dcposch:
## Android layout fixed

image

PR #583, opened 2023-12-28
contract: efficient, tx.origin-based DaimoPaymasterV2

dcposch:


packages/daimo-api/src/contract/paymaster.ts
```diff
@@ -136,10 +121,11 @@ export class Paymaster {
     // Sign paymaster for any valid Daimo account, excluding name blacklist.
     // Everyone else gets the Pimlico USDC paymaster.
     const isSponsored =
+      !chainConfig.chainL2.testnet &&
```
nalinbhardwaj, line 124: nice, with this change I think we can also add nice rate limits on paymaster spending per user / per unit time and enforce beneficiary = our address that would cleanly solve problems around paymaster griefing -- doesn't have to be this PR I guess

packages/daimo-api/src/network/bundleCompression.ts
```diff
@@ -90,14 +122,24 @@ export function compressBundle(op: UserOpHex, info: CompressionInfo) {
...
   const paymasterRegex = new RegExp(`${paymasterAddr}(.*)$`, "i");
   m = paymasterRegex.exec(op.paymasterAndData);
   if (!m) throw new Error("can't compress, bad paym.:" + op.paymasterAndData);
-  ret.push(`0x${m[1]}`); // paymaster data
+  if (m[1].length > 0) throw new Error("can't compress, has paymasterData");
 
   return concatHex(ret);
 }
+
+function compressAddr(addr: Address, nameReg: NameRegistry) {
```
nalinbhardwaj, line 137: one subtle realisation I had reading this: even though nameRegistry has upgrade keys we own we can't ever redirect any transfers even if we maliciously changed name registry because the user signs the uncompressed operation with the intended address -- nice property.

nalinbhardwaj (approved):
lgtm

Issue #366, opened 2023-10-28
Too many transfer logs on mainnet


nalinbhardwaj:
```
/opt/render/project/src/packages/daimo-api/src/cron.ts:55
Oct 28 06:32:41 PM      this.transfersQueue.push(...logs);
Oct 28 06:32:41 PM                          ^
Oct 28 06:32:41 PM  RangeError: Maximum call stack size exceeded
Oct 28 06:32:41 PM      at pipeTransfers (/opt/render/project/src/packages/daimo-api/src/cron.ts:55:25)
Oct 28 06:32:41 PM      at CoinIndexer.pipeAllTransfers (/opt/render/project/src/packages/daimo-api/src/contract/coinIndexer.ts:61:5)
```

And another call of `JSON.stringify` failing at https://github.com/daimo-eth/daimo/blob/6e09c5fc9426576612c449cff951136ad9ef0122/packages/daimo-api/src/pushNotifier.ts#L82

dcposch:
fixed

PR #584, opened 2023-12-28
api: fix spread operator stack size exceeded

nalinbhardwaj:
Prod build has been failing for the last ~12 hours due https://github.com/daimo-eth/daimo/issues/366

Doesn't affect any services since Render continues using the old successful build.


dcposch (approved):
LGTM

PR #598, opened 2024-01-04
Add pointer events wrapper around debug modal

wojtus7:
Added wrapper with dynamic state-based pointerEvents prop to make sure that debug modal doesn't prevent interactions even if invisible. 

Android:

https://github.com/daimo-eth/daimo/assets/42337257/4e9ef637-ead6-4824-967c-a4a9866a4b00

iOS: 

https://github.com/daimo-eth/daimo/assets/42337257/72da12f5-c85e-457f-bd01-3be2d417a8cd

### DIFF
--- a/apps/daimo-mobile/src/App.tsx
+++ b/apps/daimo-mobile/src/App.tsx
@@ -66,7 +66,6 @@ function AppBody() {
   useEffect(() => {
     const subscription = RNShake.addListener(() => {
       bottomSheetRef.current?.expand();
-      setIsDebugModalOpen(true);
     });
 
     return () => {
@@ -86,11 +85,7 @@ function AppBody() {
     []
   );
 
-  const onChangeIndex = (index: number) => {
-    if (index === -1) {
-      setIsDebugModalOpen(false);
-    }
-  };
+  const onChangeIndex = (index: number) => setIsDebugModalOpen(index > -1);
 
   const [sendDL] = useSendDebugLog();
 
--- a/apps/daimo-mobile/src/App.tsx
+++ b/apps/daimo-mobile/src/App.tsx
@@ -4,7 +4,7 @@ import { DefaultTheme, NavigationContainer } from "@react-navigation/native";
 import { useFonts } from "expo-font";
 import * as SplashScreen from "expo-splash-screen";
 import { StatusBar } from "expo-status-bar";
-import { useCallback, useEffect, useMemo, useRef } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { StyleSheet, View } from "react-native";
 import { GestureHandlerRootView } from "react-native-gesture-handler";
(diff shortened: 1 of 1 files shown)


apps/daimo-mobile/src/App.tsx
```diff
@@ -76,37 +78,49 @@ function AppBody() {
     (props: BottomSheetDefaultBackdropProps) => (
       <BottomSheetBackdrop
         {...props}
-        disappearsOnIndex={-1}
+        disappearsOnIndex={-0.9}
```
dcposch, line 81: ...i thought the index is an integer?

apps/daimo-mobile/src/App.tsx
```diff
@@ -76,37 +78,49 @@ function AppBody() {
...
-      <BottomSheet
-        handleComponent={ScrollPellet}
-        backdropComponent={renderBackdrop}
-        ref={bottomSheetRef}
-        index={-1}
-        snapPoints={snapPoints}
-        enablePanDownToClose
+      <View
+        style={styles.bottomSheetWrapper}
+        pointerEvents={isDebugModalOpen ? "auto" : "none"}
```
dcposch, line 103: 👍

apps/daimo-mobile/src/App.tsx
```diff
@@ -76,37 +78,49 @@ function AppBody() {
...
         pressBehavior="close"
       />
     ),
     []
   );
 
+  const onChangeIndex = (index: number) => {
+    if (index === -1) {
+      setIsDebugModalOpen(false);
+    }
```
dcposch, line 92: alternately, you can remove the `setIsDebugModalOpen(true)` above and replace this function with

```
const onChangeIndex = (index: number) => setIsDebugModalOpen(index > -1)
```

dcposch (approved):
LGTM after comments

apps/daimo-mobile/src/App.tsx
wojtus7: This index is traveling from 0 to -1 in a fluent way during modal movement, where it is interpolated between values given in snap points array.
So if snap points for 0 and -1 are 300 and 100 then -0.9 means that the backdrop should disappear when the offset is around 120 instead of strictly 100.

apps/daimo-mobile/src/App.tsx
wojtus7: Yup, but I didn't want to rely on bottom sheet logic here, assuming it can be broken and resulting in app gestures not responding.

Wanted to create something more reliable based on simple state to control it

wojtus7: But now when I'm thinking about it well, it I make it `true` and it doesn't work for real then it will be stuck in app freeze state forever ;-;

You are right it should be handled here it the same way so it rather doesn't turn on at all

PR #601, opened 2024-01-07
add metapaymaster check to slack notif

nalinbhardwaj:
still check some small balance so in case metapaymaster is out of funds we get notified still


dcposch (approved):
LGTM

PR #602, opened 2024-01-08
perf: refresh history list only on change

wojtus7:
Before it refreshed all list every time something called resync(), so every 5 seconds or even more often

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/f1634164-d480-44f6-9071-61058af42ad8

Android:

https://github.com/daimo-eth/daimo/assets/42337257/db11b53c-7741-4755-ae1f-c65b6359035c

### DIFF
--- a/apps/daimo-mobile/src/view/screen/HomeScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/HomeScreen.tsx
@@ -1,5 +1,6 @@
+import { OpStatus } from "@daimo/common";
 import Octicons from "@expo/vector-icons/Octicons";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   Dimensions,
   RefreshControl,
@@ -54,6 +55,7 @@ function HomeScreenInner({ account }: { account: Account }) {
     const showActionTimeout = setTimeout(() => setIsActionVisible(true), 1500);
     return () => clearTimeout(showActionTimeout);
   }, []);
+  const onHideSuggestedAction = () => setIsActionVisible(false);
 
   // Initialize DaimoOpSender immediately for speed.
   const keySlot = account.accountKeys.find(
@@ -71,8 +73,6 @@ function HomeScreenInner({ account }: { account: Account }) {
 
   // Outer scroll: pull to refresh
   const [refreshing, setRefreshing] = useState(false);
-
-  const onHideSuggestedAction = () => setIsActionVisible(false);
 
   const onRefresh = useCallback(async () => {
     setRefreshing(true);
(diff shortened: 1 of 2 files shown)


apps/daimo-mobile/src/view/screen/HomeScreen.tsx
```diff
@@ -44,6 +53,14 @@ function HomeScreenInner({ account }: { account: Account }) {
   const translationY = useSharedValue(0);
   const [isActionVisible, setIsActionVisible] = useState(false);
 
+  const historyStatus = useRef<CountStatus>({});
+  const [histListMini, setHistListMini] = useState(
+    <HistoryListSwipe account={account} showDate={false} maxToShow={5} />
+  );
+  const [histListFull, setHistListFull] = useState(
+    <HistoryListSwipe account={account} showDate />
+  );
```
dcposch, line 62: ive never seen this pattern of `useState` with a rendered component...

the CountStatus trick is elegant & makes sense, but maybe we can prevent HistListSwipe rerendering a different way?

eg out in HomeScreen,

```
const [account] = useAccount();
if (!account) return null;
const countStatus = countBy(...);
const accountMemo = useMemo(() => account, [countOpStatus]);
```
...then accountMemo stays reference equal until the status changes

dcposch (approved):
LGTM otherwise

apps/daimo-mobile/src/view/screen/HomeScreen.tsx
```diff
@@ -1,4 +1,5 @@
 import Octicons from "@expo/vector-icons/Octicons";
+import { countBy, isEqual } from "lodash";
```
nalinbhardwaj, line 2: in my experience before `lodash` has been a very heavy dependency / high bundle size etc. Is that the case for `react-native` bundling too? if so might recommend against using it and pulling out the `statusCount` into its own file in `../logic/`

wojtus7: Removed 🫡

apps/daimo-mobile/src/view/screen/HomeScreen.tsx
```diff
@@ -98,11 +98,23 @@ function HomeScreenInner({ account }: { account: Account }) {
...
   );
-  const histListFull = <HistoryListSwipe account={account} showDate />;
+  const histListMini = useMemo(
+    () => <HistoryListSwipe account={account} showDate={false} maxToShow={5} />,
+    [statusCountsStr]
+  );
+  const histListFull = useMemo(
+    () => <HistoryListSwipe account={account} showDate />,
+    [statusCountsStr]
+  );
```
dcposch, line 115: cc @wojtus7 , simplified a bit using `useMemo` instead of `useState`

PR #604, opened 2024-01-09
Delay focus when go to send screen from account screen

### LINKED ISSUE #541: Send screen from account page is unfocused
https://github.com/daimo-eth/daimo/assets/6984346/90e3f95c-3e3e-4157-b4ce-efd79eb9e08b

Might be fixable with using `useFocusOnScreenTransitionEnd` but we should find a better solution to the autoFocus/lagAutoFocus param in general and clean up that code change if possible.

wojtus7:
Keyboard focus was broken when going to send screen from account screen. 
It opens correctly now

Closes #541

Android:

https://github.com/daimo-eth/daimo/assets/42337257/ceddb83c-a2b9-4050-97b4-8a23a4962b57

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/54fd33eb-078c-411e-b639-7649fc9f1fdc

### DIFF
--- a/apps/daimo-mobile/src/view/screen/AccountScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/AccountScreen.tsx
@@ -111,7 +111,10 @@ function AccountScreenBody({
   const canSend = canSendTo(eAcc);
   const send = useCallback(() => {
     const recipient = addLastSendTime(account, eAcc);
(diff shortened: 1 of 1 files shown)


nalinbhardwaj (approved):
lgtm thanks

if you have ideas for removing the lag param altogether that might be a nice code cleanup too

PR #607, opened 2024-01-10
mobile: move use existing poll to intro page

nalinbhardwaj:
Protect users from deleting app and losing money.

Video: https://streamable.com/cdf6lc

### DIFF
--- a/apps/daimo-mobile/src/view/screen/onboarding/IntroPages.tsx
+++ b/apps/daimo-mobile/src/view/screen/onboarding/IntroPages.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useState } from "react";
+import { ReactNode, useEffect, useState } from "react";
 import {
   NativeScrollEvent,
   NativeSyntheticEvent,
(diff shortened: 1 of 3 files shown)


apps/daimo-mobile/src/view/screen/onboarding/IntroPages.tsx
```diff
@@ -26,6 +34,10 @@ export function IntroPages({
     setPageIndex(page);
   };
 
+  useEffect(() => {
+    if (useExistingPubKeyHex && useExistingStatus === "success") existingNext();
```
dcposch, line 38: we might want an `existing account found` page here where user can either join it or cancel
(if the account has money in it and only one key, then option to cancel can be disabled)

needs design though, future work

dcposch (approved):
LGTM

apps/daimo-mobile/src/view/screen/onboarding/IntroPages.tsx
nalinbhardwaj: I think this is a reasonable 80/20 for now

PR #610, opened 2024-01-11
Add activate offset to bottom sheet to allow back swipe

### LINKED ISSUE #551: Back swipe doesn’t work on bottom sheet
Swiping in the bottom sheet on account page does lead back, but swiping above it does — feels inconsistent

wojtus7:
Closes #551

Just like title: Allow to swipe left and right on home screen and back swipe on Account screen

iOS: (Sorry for no back interaction but it works well on physical device but doesn't seem to work at all on my simulator)

https://github.com/daimo-eth/daimo/assets/42337257/5a97c4e4-3525-4236-9433-5d0772465595

Android:

https://github.com/daimo-eth/daimo/assets/42337257/e3800d6c-7c6b-40cb-a5ed-f82d9580d54c

### DIFF
--- a/apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
+++ b/apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
@@ -1,4 +1,7 @@
-import BottomSheet, { BottomSheetBackdrop } from "@gorhom/bottom-sheet";
+import BottomSheet, {
+  BottomSheetBackdrop,
+  SCREEN_WIDTH,
+} from "@gorhom/bottom-sheet";
 import { BottomSheetDefaultBackdropProps } from "@gorhom/bottom-sheet/lib/typescript/components/bottomSheetBackdrop/types";
 import {
   ReactNode,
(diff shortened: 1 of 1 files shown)


apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
```diff
@@ -123,6 +126,8 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
         enablePanDownToClose={false}
         enableHandlePanningGesture={!disabled}
         enableContentPanningGesture={!disabled}
+        activeOffsetX={[-SCREEN_WIDTH, SCREEN_WIDTH]}
+        activeOffsetY={[-10, 10]}
```
nalinbhardwaj, line 130: is the y offset no-op? not sure I follow why we need it

wojtus7: But it was not intuitive for me as well so I tried to do it with only `activeOffsetX` and was wondering why all bottom sheet doesn't work 🫠

nalinbhardwaj: sounds good, looks like other people do similar numbers too: https://github.com/search?q=activeOffsetX%3D+&type=code

nalinbhardwaj (approved):

PR #627, opened 2024-01-16
web: revert payment links look like invites

nalinbhardwaj:
the app doesn't have payment link invites enabled yet so anyone landing on this page will be confused

just ran `git checkout 13796c2 apps/daimo-web/`, can revert when we release.


nalinbhardwaj:
superseded by #628

PR #631, opened 2024-01-17
Add maximum font size to handle XL font on iOS

### LINKED ISSUE #449: Handle accessibility settings
Currently, XL font > broken layout.

Proposed fix: Limit maximum font size to something reasonable that doesn’t break render globally: one place to override everything, tabs, home, bubble etc

wojtus7:
Add maximum font size and don't allow font scaling in some places.
Maybe we should consider all <Text> being imported from text.tsx file instead of React Native but in this PR I added necessary prop to all instances without risk of breaking anything

Closes #449

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/043347b1-7848-434b-a162-4806a55c0c4a

Android:

https://github.com/daimo-eth/daimo/assets/42337257/215e0d3a-d11e-4b19-bb4c-cf2c89ace5b6

### DIFF
--- a/apps/daimo-mobile/src/view/TabNav.tsx
+++ b/apps/daimo-mobile/src/view/TabNav.tsx
@@ -103,6 +103,7 @@ function getTabOptions(
     tabBarIconStyle: {
       alignItems: "center",
     },
+    tabBarAllowFontScaling: false,
   };
   switch (route.name) {
     case "DepositTab":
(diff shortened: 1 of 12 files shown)


nalinbhardwaj (changes requested):
> Maybe we should consider all being imported from text.tsx file instead of React Native but in this PR I added necessary prop to all instances without risk of breaking anything

Good call, I like that so we don't have to worry about this issue in future.

Want to refactor to export a `Text` or `DaimoText` component instead that just adds `maxFontSizeMultiplier` to RN's `Text`?

`allowFontScaling={false}` seems rare enough that we don't need to do the same there IMO.

wojtus7:
@nalinbhardwaj fixed and created DaimoText that injects max font scale. The only <Text> component left is the one inside ContactBubble but it is more like an Icon than real text so I let it be not to confuse it with different text in the app.

Also left <TextInput> because it would additional layer of passing 'ref' up the react tree so I don't think this is necessary to do for now only for one prop

nalinbhardwaj (approved):
lgtm

PR #646, opened 2024-01-24
onboarding: fix key load or create race

nalinbhardwaj:
unify key load or creation to prevent race condition between createAccount and useExisting.

tested by running through onboarding > delete device a handful of times.

### DIFF
--- a/apps/daimo-mobile/src/action/useCreateAccount.ts
+++ b/apps/daimo-mobile/src/action/useCreateAccount.ts
@@ -10,7 +10,7 @@ import { DaimoChain } from "@daimo/contract";
 import { useEffect } from "react";
 
 import { ActHandle, useActStatus } from "./actStatus";
-import { useLoadOrCreateEnclaveKey } from "./key";
+import { DeviceKeyStatus } from "./key";
 import { createEmptyAccount } from "../logic/account";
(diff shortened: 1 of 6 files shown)


apps/daimo-mobile/src/action/useCreateAccount.ts
```diff
@@ -19,12 +19,11 @@ import { defaultEnclaveKeyName, useAccount } from "../model/account";
 export function useCreateAccount(
   name: string,
   inviteLink: DaimoLinkInvite | DaimoLinkNoteV2 | undefined,
-  daimoChain: DaimoChain
+  daimoChain: DaimoChain,
+  keyStatus: DeviceKeyStatus
```
dcposch, line 23: i really want to get deeper app logic out of react components 

(and specifically, out of our hooks)

we can use `AccountManager` (or a new `OnboardingManager`) or similar to own this logic instead

OK for now, and it's not rly your fault--hooks just inherently tend toward spaghetti

dcposch: fine for now.

fixing this is a deeper problem. noted here:

https://github.com/daimo-eth/daimo/issues/654

dcposch (approved):

Issue #569, opened 2023-12-20
Show error page for broken deeplinks


dcposch:
## Goal

- User clicks a deeplink, opening Daimo.
- Deeplink is broken for any reason (typo, indexer issue, old version of the app, ...)

Current behavior: nothing happens. App ignores the deeplink.
**Goal: always respond to link. If necessary, show a helpful error.**

## Proposed solution

- Recognized link - nav as normal, then show error.
  - Account. Open the account screen. Potentially show account not found state.
  - Payment link. Open the note screen. Potentially show error.
  - Request link. Open the send screen. Potentially show error.
- Unrecognized link - show error modal bottomsheet in home tab. Show unrecognized link error. Link to app store to get latest version of app.

Design TBD.

michelle-chiu:
<img width="936" alt="Screenshot 2024-01-05 at 6 11 00 PM" src="https://github.com/daimo-eth/daimo/assets/5884103/6ffa73ab-ebcf-42e0-b2c3-772f57db5b08">

Designs can be found here: https://www.figma.com/file/wBBNeEzlsObaE8A8i9iVf7/Mobile-App?type=design&node-id=2869%3A12606&mode=design&t=oOYJDxxTRRgghFsc-1

michelle-chiu:
<img width="271" alt="Screenshot 2024-01-05 at 6 23 41 PM" src="https://github.com/daimo-eth/daimo/assets/5884103/c0065114-47fe-4fa3-9a62-d15bb3b7d499">

Also designed a bottom sheet version if we want to go in that direction instead

dcposch:
done in #649

follow up to fix backdrop #691

PR #649, opened 2024-01-25
Handle not working deeplinks

wojtus7:
It should handle broken deeplinks - both in screens and if link is utterly broken using bottom sheet.

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/22b7aea7-6f39-4d74-98c3-57d9886ff937

Android:

https://github.com/daimo-eth/daimo/assets/42337257/f49b678a-93fe-42de-9e94-f3ff7c7a804d

### DIFF
--- a/apps/daimo-mobile/src/view/TabNav.tsx
+++ b/apps/daimo-mobile/src/view/TabNav.tsx
@@ -11,11 +11,13 @@ import {
   createNativeStackNavigator,
 } from "@react-navigation/native-stack";
 import {
-  createStackNavigator,
+  StackCardInterpolationProps,
+  StackCardStyleInterpolator,
   TransitionPresets,
+  createStackNavigator,
 } from "@react-navigation/stack";
 import { useEffect, useState } from "react";
-import { Platform, Animated } from "react-native";
+import { Animated, Platform } from "react-native";
 import { EdgeInsets, useSafeAreaInsets } from "react-native-safe-area-context";
 
 import { AccountScreen } from "./screen/AccountScreen";
@@ -84,23 +86,13 @@ export function TabNav() {
 
   if (!isOnboarded) return <OnboardingScreen {...{ onOnboardingComplete }} />;
 
-  function forModalPresentationIOS({
-    current, // Animated node representing the progress value of the current screen
-    next, // Animated node representing the progress value of the next screen.
-    inverted, // Animated node representing multiplier when direction is inverted (-1 - inverted, 1 - normal).
-    layouts: { screen }, // Layout measurements of the whole screen.
-  }: {
-    current: {
-      progress: Animated.AnimatedInterpolation<number>;
-    };
-    next?: {
-      progress: Animated.AnimatedInterpolation<number>;
-    };
-    inverted: Animated.AnimatedInterpolation<-1 | 1>;
-    layouts: {
-      screen: { width: number; height: number };
-    };
-  }) {
+  // Error modal slides up from the bottom, greying out the app below.
+  const errorModalInterpolatorIOS: StackCardStyleInterpolator = ({
+    current,
+    next,
+    inverted,
+    layouts: { screen },
+  }: StackCardInterpolationProps) => {
     const progress = add(
       current.progress.interpolate({
         inputRange: [0, 1],
@@ -135,7 +127,7 @@ export function TabNav() {
       },
       overlayStyle: { opacity: overlayOpacity },
     };
-  }
+  };
 
   return (
     <MainStack.Navigator initialRouteName="MainTabNav">
@@ -162,7 +154,7 @@ export function TabNav() {
             ...TransitionPresets.ModalPresentationIOS,
             detachPreviousScreen: false,
             gestureResponseDistance: WINDOW_HEIGHT,
-            cardStyleInterpolator: forModalPresentationIOS,
+            cardStyleInterpolator: errorModalInterpolatorIOS,
           }}
         />
       </MainStack.Group>
--- a/apps/daimo-mobile/src/view/TabNav.tsx
+++ b/apps/daimo-mobile/src/view/TabNav.tsx
@@ -1,5 +1,6 @@
 import { assertUnreachable } from "@daimo/common";
 import Octicons from "@expo/vector-icons/Octicons";
+import { WINDOW_HEIGHT } from "@gorhom/bottom-sheet";
 import {
   MaterialTopTabNavigationOptions,
   createMaterialTopTabNavigator,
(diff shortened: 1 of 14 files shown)


dcposch:
Fixes #569

michelle-chiu:
Here's my copy suggestion:

**Title:** This link is unsupported
**Description:** Check if you have an old version of the app or if there are any errors in your URL
**Primary Button:** [DOWNLOAD LATEST VERSION]
**Secondary Button:** [BACK TO HOMEPAGE]

apps/daimo-mobile/src/view/shared/error.tsx
```diff
@@ -16,3 +22,86 @@ export function ErrorRowCentered({ error }: { error: { message?: string } }) {
...
+  const goAppStore = () => {};
+  console.log(
+    `[ERROR] rendering ${error?.message || "Unknown message"}`,
+    error
+  );
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.closeCircle}>
+        <Image source={Close} style={styles.closeIcon} />
```
dcposch, line 48: recommend using the same octicon we use for close elsewhere--see ScreenHeader for example

you can still make the color red if there's a good reason to do so. otherwise, i'd default to same color ^

packages/daimo-common/src/daimoLink.ts
```diff
@@ -131,9 +139,20 @@ function formatDaimoLinkInner(link: DaimoLink, linkBase: string): string {
     case "tag": {
       return `${linkBase}/t/${link.id}`;
     }
+    case "error": {
+      return `${linkBase}/error`;
+    }
```
dcposch, line 144: i don't think this needs to be its own link type?

what would be the point of deeplinking to https://daimo.com/l/error ?

packages/daimo-common/src/daimoLink.ts
```diff
@@ -163,61 +182,64 @@ function parseDaimoLinkInner(link: string): DaimoLink | null {
...
+      if (parts.length !== 4) return defaultError;
       const recipient = parts[1];
       const dollarNum = parseFloat(zDollarStr.parse(parts[2]));
-      if (!(dollarNum > 0)) return null;
+      if (!(dollarNum > 0)) return defaultError;
       const dollars = dollarNum.toFixed(2) as DollarStr;
       const requestId = `${BigInt(parts[3])}` as BigIntStr;
 
-      if (dollars === "0.00") return null;
+      if (dollars === "0.00") return defaultError;
```
dcposch, line 203: tbh i think this whole file (`daimoLink.ts`) was fine before-- we just need handling in the event `parseDaimoLink` returns null

we can move the `exp+daimo` development links check: that's the only type of link i can think of we'd want to *silently* ignore

in all other cases, if a link is passed to the app as a deeplink:
- call parseDaimoLink
- if we get a parsed result, display
- otherwise, display the error screen

dcposch (changes requested):
Looks close--left comments

apps/daimo-mobile/src/view/shared/error.tsx
wojtus7: I used png from figma because it looked ever so slightly different but I gotcha that in terms of code clarity using the octicon component makes so much more sense 🫡

apps/daimo-mobile/src/view/TabNav.tsx
```diff
@@ -73,6 +71,105 @@ export function TabNav() {
...
+  useEffect(() => {
+    // This is a latch: if we clear the account, go back to onboarding.
+    if (isOnboarded && account == null) setIsOnboarded(false);
+  }, [isOnboarded, account]);
+  // Stay onboarding till it's complete.
+  const onOnboardingComplete = () => setIsOnboarded(true);
+
+  if (!isOnboarded) return <OnboardingScreen {...{ onOnboardingComplete }} />;
+
+  function forModalPresentationIOS({
```
dcposch, line 87: recommend adding a short comment explaining what this does. in particular, what's `current`, `next`, and `inverted`?

apps/daimo-mobile/src/view/shared/error.tsx
```diff
@@ -16,3 +23,89 @@ export function ErrorRowCentered({ error }: { error: { message?: string } }) {
...
+  const goAppStore = () => {};
+  console.log(
+    `[ERROR] rendering ${error?.message || "Unknown message"}`,
+    error
+  );
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.closeCircle}>
+        <Octicons name="x" size={74} color={color.danger} />
```
dcposch, line 49: 👍

apps/daimo-mobile/src/view/TabNav.tsx
```diff
@@ -73,6 +71,105 @@ export function TabNav() {
...
+        }}
+      >
+        <MainStack.Screen
+          name="LinkErrorModal"
+          component={ErrorScreen}
+          options={{
+            ...TransitionPresets.ModalPresentationIOS,
+            detachPreviousScreen: false,
+            gestureResponseDistance: WINDOW_HEIGHT,
+            cardStyleInterpolator: forModalPresentationIOS,
```
dcposch, line 165: looks reasonable to me

i'm curious why this is implemented differently from the shake-gesture Send Debug Log (which is also a bottom-sheet popup that can appear on any screen at any time)

dcposch (approved):
LGTM after comments

apps/daimo-mobile/src/view/TabNav.tsx
wojtus7: When I created Send Debug Log there was no stack navigator installed and it could be opened just from one source: The shake gesture. I think it was reasonable not to rewrite some part of navigation to add it even thou it can be opened just from one listener. Thats why I decided to use Bottom Sheet that was already in the app and use it for that purpose.

This bad deeplink can be called from anywhere in the app (QRScreen, Home Screen and maybe some more places in the future) so I kinda wanted to have it in the navigation to make it reasonable to use from anywhere. And for that I needed to heavily modify the navigation and install stack navigator.

Now with navigation rewritten to support modals like this I think this is a great idea to move Send Debug Log modal to work exactly the same as this new Bad Deeplink modal.

Maybe create ticket for it and make it as follow up PR?

apps/daimo-mobile/src/view/TabNav.tsx
wojtus7: Added 👌

apps/daimo-mobile/src/view/TabNav.tsx
dcposch: yeah was just a question! looks good to me

PR #651, opened 2024-01-26
mobile: do not double handle initalURL

### LINKED ISSUE #650: initial URL should not be re-read after onboarding
flow: close app (including not in foreground) > open app with link > go through onboarding > it re-opens the note page, now you have a pending tx in your recents

nalinbhardwaj:
https://streamable.com/xxf73e

Fixes #650 : if the app is launched by an initialURL during onboarding, it will double trigger action to it which looks bad after onboarding from payment links


dcposch (approved):
LGTM

feels a bit hack--i'm not a fan of module globals-- but fine for now and unblocks us

dcposch (approved):
LGTM

feels a bit hack--i'm not a fan of module globals-- but fine for now and unblocks us

PR #667, opened 2024-01-30
api: bump shovel limit

nalinbhardwaj:
Couple server fails in the last day with `Query read timeout` https://logs.betterstack.com/team/183561/tail?a=1706558645199434-104000491&rf=now-60m&q=syslog.host%3D%22daimo-api-prod%22

hotfix for now, in future we should add a retry loop

### DIFF
--- a/packages/daimo-api/src/shovel/watcher.ts
+++ b/packages/daimo-api/src/shovel/watcher.ts
@@ -9,9 +9,9 @@ interface indexer {
 
 const dbConfig: ClientConfig = {
   connectionString: process.env.SHOVEL_DATABASE_URL,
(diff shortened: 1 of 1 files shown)


ryandotsmith:
I'm not able to see the logs in better stack. Do you have specific queries that are timing out? I wonder if we are missing an index.

ryandotsmith (approved):
i'd like to find the root cause, but in a pinch, this lgtm!

dcposch (approved):

dcposch:
@ryandotsmith added you:

https://logs.betterstack.com/team/183561/tail

PR #668, opened 2024-01-31
Use % instead of numbers to handle resize of Mac app

### LINKED ISSUE #659: macOS bottom sheet doesn’t appear
Fix bottom sheet interactions, it appears we’re doing some screen size based math which might be putting it below the window

wojtus7:
Closes #659

Handle bottom sheet so it works even after resizing on Mac OS (Still looks kinda strange while transition though but it usable)

Mac OS:

https://github.com/daimo-eth/daimo/assets/42337257/a8222bf8-b421-4595-a5e3-a6b043fd830d

After tests android and iOS seems to working same as before change

### DIFF
--- a/apps/daimo-mobile/src/view/shared/useSwipeUpDown.tsx
+++ b/apps/daimo-mobile/src/view/shared/useSwipeUpDown.tsx
@@ -25,13 +25,6 @@ export function useSwipeUpDown({
 }) {
   const [isBottomSheetOpen, setIsOpen] = useState(false);
 
-  // Dimensions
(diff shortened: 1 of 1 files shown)


nalinbhardwaj (approved):
lgtm; nice cleanup

nalinbhardwaj:
One more related question from looking at the video: is it possible to set the app window size on Mac to be similar (proportionally) to phone screen dimensions? The app looks kind of wack at other sizes anyway (buttons etc.)

If it’s not possible to lock the size from resizing maybe it’s possible to set the initial window size?

PR #673, opened 2024-02-04
Remove react navigation from bottom sheet and use reanimated for smoother experience

### LINKED ISSUE #660: Create Reanimated driven navigation in bottom sheet
Instead of using react navigation we can rewrite how history op screen works to use reanimated connected with bottom sheet position's shared value to get more responsive interface 🙏

wojtus7:
Remove react navigation from bottom sheet and use reanimated for smoother experience.

Closes #660

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/98f3606a-78b1-4036-9a32-037920a0c692

Android:

https://github.com/daimo-eth/daimo/assets/42337257/971b3993-3a25-4aed-9003-0caaeb9c3d0c

### DIFF
--- a/apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
+++ b/apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
@@ -15,7 +15,9 @@ import {
 } from "react";
 import { Dimensions, StyleSheet, ViewProps } from "react-native";
 import Animated, {
+  runOnJS,
   useAnimatedProps,
+  useAnimatedReaction,
   useAnimatedStyle,
   useSharedValue,
   withTiming,
@@ -72,6 +74,7 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
       screenDimensions.height - ins.top - ins.bottom - tabBarHeight
     );
     const snapPoints = useSharedValue([posYMini, posYFull - 1, posYFull]);
+    const nextSnapPoint = useSharedValue(-1);
 
     useEffect(() => {
       const maxHeightOffset = screenDimensions.height - ins.top - ins.bottom;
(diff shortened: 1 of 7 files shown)


apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
```diff
@@ -81,22 +80,32 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
...
     const sheetExpand = () => {
       if (animatedIndex.value === 1) {
         bottomRef.current?.snapToIndex(2);
       }
-      setSnapPoints([posYMini, 450, posYFull]);
-      if (animatedIndex.value === 0) {
-        bottomRef.current?.snapToIndex(1);
-      }
+      snapPoints.value = [posYMini, 450, posYFull];
+      // skip one frame to let snapPoints addapt
```
dcposch, line 92: what do you mean by by "let snapPoints adapt"

apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
```diff
@@ -81,22 +80,32 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
...
+      // without it it works perfectly fine
+      // but the backdrop is gray right away without smooth transition
+      setTimeout(() => {
+        if (animatedIndex.value === 0) {
+          bottomRef.current?.snapToIndex(1);
+          historyOpOpacity.value = 1;
```
dcposch, line 98: what's the purpose of this line? 
it looks like we're overwriting to `withTiming(1)` immediately after

apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
```diff
@@ -125,15 +134,12 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
       onShowFull?.();
     };
 
-    const nav = useNav();
```
dcposch, line 128: nice cleanup

dcposch (approved):
LGTM after comments

nice UX upgrade!

remove the harcoded `setTimeout(..., 20)` if possible

apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
wojtus7: When I call `historyOpOpacity.value = 1;` just before `withTiming(1)` I basically skip the animation of the withTiming because the value is already there.

The `historyOpOpacity.value = 1;` is inside `if` because I wanted to skip the animation conditionally but now when I look at it maybe I will use more undestandable syntax:
```
historyOpOpacity.value = animatedIndex.value === 0 ? 1 : withTiming(1)
```

apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
wojtus7: When I change the `snapPoints.value` it changes but with slight delay, so when I tried to call `snapToIndex` just after this line without timeout it still snapped to old snapPoints 😔

I will try to make it somehow else so it doesnt use hardcoded timeout

wojtus7: Used useAnimatedReaction instead and seems to work fine 🙏

PR #680, opened 2024-02-06
daimo-api: update shovel watcher to use latest shovel schema

ryandotsmith:
The latest version of shovel exposed a view that callers can use to see where Shovel is at in the chain.

https://github.com/indexsupply/code/pull/232

### DIFF
--- a/packages/daimo-api/src/shovel/watcher.ts
+++ b/packages/daimo-api/src/shovel/watcher.ts
@@ -23,6 +23,7 @@ const poolConfig: PoolConfig = {
 
 export class Watcher {
   private latest = chainConfig.chainL2.testnet ? 8750000n : 5700000n;
+  private srcName = chainConfig.chainL2.testnet ? "base-goerli" : "base";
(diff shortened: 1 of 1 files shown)


packages/daimo-api/src/shovel/watcher.ts
```diff
@@ -107,13 +108,8 @@ export class Watcher {
...
   async getShovelLatest(): Promise<bigint> {
     const result = await this.pg.query(
-      `
-      select max(num) as num
-      from shovel.task_updates
-      where chain_id = $1
-      and backfill = false;
-    `,
-      [chainConfig.chainL2.id]
+      `select num from shovel.latest where src_name = $1 `,
```
nalinbhardwaj, line 111: I think the `src_name` filter is not necessary since we've separated testnet/stage and mainnet/prod shovels?

nalinbhardwaj (approved):
lgtm; seems to work fine now on testnet APIs

Issue #683, opened 2024-02-08
Incident: API down


nalinbhardwaj:
## Summary

**Daimo API was down for 4h on Feb 8**. It appears to be triggered by (intermittently) very slow postgres queries due to an issue in Render, which revealed bugs in our indexing system. What caused the slowness is unclear, but Render's own site was also partly down during this time, showing intermittent 404 pages.

<img width="500" alt="image" src="https://github.com/daimo-eth/daimo/assets/169280/0cb6de31-1e4f-4e8e-9724-c9c4ba13653e">

## Timeline

**2024-02-08 Pacific time**

- **11:11am** - [payment link create transaction](https://basescan.org/tx/0x97168deec27612e11a369e423088a487457a452a2256cc7b66b940f9cd494463)
- **11:11** - server down
- **12:00** - attempted hotfix. deploy fails due to query read timeouts. potentially unrelated bug.
- **12:20** - investigating
- **12:45** - try reducing batch size https://github.com/daimo-eth/daimo/pull/686/commits/2567f9790590d556091ebd7bb53bc43b62cf9500
- **13:00** - the error from #624 reoccurs. issue revealed by smaller batch size. fixing. 
- **13:55** - fix notelog ordering #686 , deploy. https://github.com/daimo-eth/daimo/pull/686/commits/383151814586813e7993fe3d9285f7e7fae3f008
- **14:00** - still getting 10s query timeouts during api server startup.
- **~14:30** - upgraded postgres to larger instance, per Ryan Smith's recommendation.
- **~14:45** - upgraded app server to larger instance, based on Nalin's observation that memory use was high > we're possibly limited by node-pg trying to use disk.
- **15:15** - hotfix: api: 20s query timeout, simplify transfers query https://github.com/daimo-eth/daimo/pull/686/commits/8a70b5542f30fc9f7af67bfc2ce05e088f0725da
- **~15:30** - **still failing due to intermittent 60s (!) postgres queries. we see lots of 404s in the render webapp.** intermittent-- 404s typically go away on refresh. at this point, we're pretty sure there's a render issue, potentially a networking issue between our app server and db server.
- **15:59** - hotfix: retry with exponential backoff for queries. https://github.com/daimo-eth/daimo/commit/d2e26436c4cf1e117a3be37c0d33334e5d7ffaa1
- **16:09** - **back up, api healthy**

### Query timeout

**Issue appeared once, a week ago, and solved itself via auto restart.** Followup at the time: #667 

[...]

PR #684, opened 2024-02-08
api: hotfix speedup query reads

nalinbhardwaj:
fixes two potential query read timeout for #683

### DIFF
--- a/apps/daimo-mobile/app.config.js
+++ b/apps/daimo-mobile/app.config.js
@@ -1,7 +1,7 @@
 const IS_DEV = process.env.DAIMO_APP_VARIANT === "dev";
 
-const VERSION = "1.5.1";
-const BUILD_NUM = 103;
+const VERSION = "1.5.2";
+const BUILD_NUM = 105;
 
 export default {
   owner: "daimo",


packages/daimo-api/src/contract/coinIndexer.ts
```diff
@@ -63,9 +63,7 @@ export class CoinIndexer {
           v as "value"
         from transfers
         where block_num >= $1
-        and block_num <= $2
-        and chain_id = $3
-        and log_addr = $4;
```
nalinbhardwaj, line 68: checked `select
          block_num,
          block_hash,
          tx_hash,
          tx_idx,
          log_idx,
          log_addr,
          f as "from",
          t as "to",
          v as "value"
        from transfers
        where log_addr <> '\x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913';` is empty

nalinbhardwaj: and same for chain ID

nalinbhardwaj:
idk why it closed it, I pushed the commit to prod

PR #688, opened 2024-02-09
Improved bottom sheet animations 

wojtus7:
Improved bottom sheet animations to be smoother 
Unfortunately I was forced to change the code inside bottom sheet library so to make it possible I installed patch-package.

To make the changes run `npm install` in daimo-mobiles and changes should be applied automatically during `postinstall` script

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/f1eca8ce-998c-4761-9b0a-86014b146bfe

Android:

https://github.com/daimo-eth/daimo/assets/42337257/72f7626b-fef7-414e-beac-c608c1481d4a

### DIFF
--- a/apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
+++ b/apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
@@ -15,7 +15,9 @@ import {
 } from "react";
 import { Dimensions, StyleSheet, ViewProps } from "react-native";
 import Animated, {
+  runOnJS,
   useAnimatedProps,
+  useAnimatedReaction,
   useAnimatedStyle,
   useSharedValue,
   withTiming,
@@ -72,6 +74,7 @@ export const SwipeUpDown = forwardRef<SwipeUpDownRef, SwipeUpDownProps>(
       screenDimensions.height - ins.top - ins.bottom - tabBarHeight
     );
     const snapPoints = useSharedValue([posYMini, posYFull - 1, posYFull]);
+    const nextSnapPoint = useSharedValue(-1);
 
     useEffect(() => {
       const maxHeightOffset = screenDimensions.height - ins.top - ins.bottom;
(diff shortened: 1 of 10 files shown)


apps/daimo-mobile/src/view/shared/SwipeUpDown.tsx
```diff
@@ -221,3 +259,10 @@ const styles = StyleSheet.create({
     backgroundColor: color.white,
   },
 });
+
+const ANIMATION_CONFIG = {
```
nalinbhardwaj, line 263: hmm is my understanding correct that this is what changes the bounciness (and the rest of the PR is the change for transitions of content inside in the bottom sheet?). At that point, our bottom sheet would feel as playful to move as Apple's but the transitions inside would be as-before?

If so, for code quality / maintainability I would recommend just making the bounciness change and leave the in-sheet content as is. Apologies for the extra work, didn't realize it would be require such a deep change.

wojtus7: Unfortunately if we change config inside our components it triggers rerender of bottom sheet and calls some useEffects inside of the library and as a result the bottom sheet tries to close after config change and it's a constant struggle with auto closing modal. 

I tried to turn off some useEffects that trigger auto close inside the library but even thou it looked like it worked, I am afraid that it wont trigger in some edge cases we want it to. 

This way or another some changes inside the library are necessary to achieve multiple animation configs 😞

patches/@gorhom+bottom-sheet+4.6.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx b/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
+index f20e3dc..ba5d095 100644
+--- a/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
++++ b/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
+@@ -721,7 +721,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
+         /**
+          * force animation configs from parameters, if provided
+          */
+-        if (configs !== undefined) {
++        if (configs !== undefined || animatedNextPositionIndex.value === 0) {
```
dcposch, line 10: is there any other way we could solve this?
i'd prefer to avoid `patch-package`

dcposch:
extracted #701

PR #693, opened 2024-02-10
web: fix next build OOM for web service

nalinbhardwaj:


apps/daimo-web/package.json
```diff
@@ -31,7 +31,7 @@
   },
   "scripts": {
     "dev": "next dev -p 3001",
-    "build": "next build",
+    "build": "NODE_OPTIONS=\"--max-old-space-size=1536\" next build",
```
dcposch, line 34: 👀

dcposch (approved):
LGTM if this fixes it

any idea what changed to make `next build` more memory intensive?

apps/daimo-web/package.json
dcposch: this actually reduces the max heap
from the Render error before, looks like it defaulted to 2GB = machine memory size
> 254067 ms: Scavenge (reduce) 2000.3 (2074.4) -> 1999.8 (2074.4) MB, 21.43 / 0.00 ms  (average mu = 0.262, current mu = 0.291) allocation failure;

dcposch:
^ see above, trying two more things:
- Increase Render instance size
- Use Vercel

Issue #633, opened 2024-01-17
Bottom Sheet design improvements


nalinbhardwaj:
## Aim

@michelle-chiu noted that the bottom sheet in Apple Maps is fun to play with. On the other hand, ours really doesn't feel that way..

Since it seems like bottom sheet is becoming a repeating pattern in our app (Home screen, Account / Profile screen, (potentially) Invite tab...) it seems worth investing some time into making it feel really nice.

Also just gives users something to fidget with aimlessly. Liam mentioned he will just open the app to look at his balance from time to time.

## Fixes

1. On the eng side, low hanging fruit: add a small amount of "bounciness" to the bottom sheet. Apple's bounces lightly on open/close in a very satisfying way (I've made the background grayer than it usually is in our app video to make it more visible):

https://github.com/daimo-eth/daimo/assets/6984346/d0ae67ee-8043-45c6-ab82-086931c3652d

https://github.com/daimo-eth/daimo/assets/6984346/bc614f60-de99-4b13-b5bd-904ac6836d4b

Make ours the same amount of bouncy. Maybe its possible to increase the frame rate as well and add a light drop shadow at the top of the bottom sheet?

2. Design task:
    - Apple uses color contrasts really nicely. We could color our home screen light gray, ivory or cream (like the video above) to contrast. We should also add a drop shadow to the bottom sheet top (Apple, Uber and all others seem to do it..)
    - Apple gives you space to hold and drag it up and down at different places, we currently squash rows together so not as much of this?

dcposch:
Part 1 ready for @wojtus7 

Part 2 needs design cc @michelle-chiu

michelle-chiu:
@wojtus7 - I'll get around to designs over the next day or so and will drop updates here once I have them!

michelle-chiu:
Excuse the missing bottom nav for now (will fix tomorrow), but the bouncy animation is illustrated here:

https://github.com/daimo-eth/daimo/assets/5884103/95e5a81b-d3c3-4c2c-ab1e-fe588c829285

dcposch:
done in #701

PR #701, opened 2024-02-12
mobile: simplified bottom-sheet bounce

dcposch:
## Summary

- Uses the bounce from #688, but without changing the inner content
- (The video in #633 suggested changing the direction the inner content slides too--but added too much engineering complexity.)

## Screenshot

https://github.com/daimo-eth/daimo/assets/169280/8ea046c1-9e3e-45d7-a136-ae406c8c836d


nalinbhardwaj (approved):
lgtm

PR #713, opened 2024-02-14
refactor swipe up/down 

nalinbhardwaj:
tested on iOS

closes https://github.com/daimo-eth/daimo/issues/717

### DIFF
--- a/apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
@@ -48,9 +48,11 @@ type Props = NativeStackScreenProps<
   "BottomSheetHistoryOp"
 >;
 
-export const ChangeBottomSheetSnapsContext = createContext(
-  (snaps: 2 | 3) => {}
-);
+// Allows the HistoryOpScreen to change the bottom sheet snap point count.
+// This allows the bottom sheet to be dismissed when the user exits the detail
+// screen and only display the half screen snap point when the user is on the
+// detail screen.
+export const SetBottomSheetSnapPointCount = createContext((snaps: 2 | 3) => {});
 
 export function HistoryOpScreen(props: Props) {
   const Inner = useWithAccount(HistoryOpScreenInner);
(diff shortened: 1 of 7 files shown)


apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
```diff
@@ -42,22 +43,31 @@ import {
...
 
-type Props = { account: Account; op: DisplayOpEvent };
+type Props = NativeStackScreenProps<
+  ParamListBottomSheet,
+  "BottomSheetHistoryOp"
+>;
 
-export const ToggleBottomSheetContext = createContext((expand: boolean) => {});
+export const ChangeBottomSheetSnapsContext = createContext(
+  (snaps: 2 | 3) => {}
```
dcposch, line 52: what does this do?

apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
```diff
@@ -66,12 +76,17 @@ function HistoryOpScreenInner({ account, op }: Props) {
...
+  const nav = useNav();
+
   return (
     <View style={ss.container.screen}>
       <ScreenHeader
         title="Transfer"
         onExit={() => {
-          toggleBottomSheet(false); // Collapse to small height
+          if (nav.canGoBack()) {
+            changeBottomSheetSnaps(3);
```
dcposch, line 87: it feels weird for the HistoryOpScreen (inner component) to have to change the SwipeUpDown (outer component) this way

if i understand correctly:
- SwipeUpDown always has 2 snap points (~25% open and 100% open)
- in some situations, it has a third (half screen)
- `changeBottomSheetSnaps` is how we're toggling this

dcposch (approved):
LGTM. Left comments.

I did prefer the swipe-down gesture to close the Transfer page, but bug-free is more important.

apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
nalinbhardwaj: added a comment:
> Allows the HistoryOpScreen to change the bottom sheet snap point count.
> This allows the bottom sheet to be dismissed when the user exits the detail
> screen and only display the half screen snap point when the user is on the
> detail screen.

apps/daimo-mobile/src/view/screen/HistoryOpScreen.tsx
nalinbhardwaj: yep that's correct -- when on the list screen we want to show three snap points with the lowest being the default dismissed version, when on the detail op screen we want to show two snap points with the lower being the middle "active" (with changed background) version

nalinbhardwaj: not super happy with the ugliness either but best I could come up with.

Issue #702, opened 2024-02-12
Website rebrand


dcposch:
## Goal

**Rebrand https://daimo.com**. NextJS, largely static site, few code changes, mostly style changes.

### Before > after

<img width="1413" alt="image" src="https://github.com/daimo-eth/daimo/assets/169280/bbf76bf9-cc06-403c-ab88-e63004091ffe">

Code: see `apps/daimo-web` 
New design: see Figma mocks below

## Approximate steps

- [ ] Clone repo. Follow quickstart to run `daimo-web` locally, pointing to our hosted testnet API.
- [ ] Update homepage. (See https://daimo.com/ ) See mocks below. Prioritize clean, simple code. If anything in mocks is pulling you in direction of greater complexity, we can discuss in chat.

michelle-chiu:
<img width="1605" alt="Screenshot 2024-02-12 at 11 49 56 AM" src="https://github.com/daimo-eth/daimo/assets/5884103/528f19f6-ae85-4e3c-a81f-3e02f7267517">

Figma link can be found here: https://www.figma.com/file/KKzdrpm6obI7IjmhgKG3dz/Branding?type=design&node-id=461%3A20942&mode=design&t=cybmVWULzkRpDJ3p-1

michelle-chiu:
On load animations: 

Desktop:
https://github.com/daimo-eth/daimo/assets/5884103/00bfe890-97fb-41b0-a468-8daf30c4b0b9

Mobile: 
https://github.com/daimo-eth/daimo/assets/5884103/1503e21b-dcf7-4319-ae51-2740e8335356

michelle-chiu:
Final designs can be found in this new Figma file: 
https://www.figma.com/file/3HKkiP8QQ5CnBq1BH3mbzz/Website?type=design&node-id=1%3A82&mode=design&t=kktRONK2LiswLFfL-1

<img width="1196" alt="Screenshot 2024-02-19 at 12 22 02 PM" src="https://github.com/daimo-eth/daimo/assets/5884103/3210bff2-c434-4aa6-b85d-08e3cf7c347c">

ItsApex:
I can hop on this 

here are few projects 
https://linkcollect.io/
https://zedit.io/

And there are few apps we are working in based out of UK
Would love to craft this one as well :)

dcposch:
@ItsApex thanks for the offer. Found someone!

dcposch:
@ItsApex thanks for the offer. Found someone already!

[...]

PR #740, opened 2024-02-22
Daimo Design Refresh

pfista:
Draft PR to track progress on implementing new designs specified in #702


apps/daimo-web/src/components/PixelBackground.tsx
```diff
@@ -0,0 +1,6929 @@
+export function PixelBackground(
```
dcposch, line 1: hmmmmm this will create a huge bundle size, right?

dcposch: recommend linking to asset image

apps/daimo-web/tailwind.config.js
```diff
@@ -11,13 +11,24 @@ module.exports = {
       yellow: "#FFDC62",
       success: "#0CA01B",
       primary: "#007aff",
-      primaryLight: "#2888f7",
+      primaryLight: "#144B44",
```
dcposch, line 14: are we not using `primary` anywhere? otherwise it should change to green too

dcposch: all good, i can harmonize the color constants with the in-app ones in post

dcposch (approved):
LGreatTM

apps/daimo-web/src/app/globals.css
```diff
@@ -15,6 +15,7 @@ html {
 body {
   overscroll-behavior-y: none;
   position: relative;
+  min-width: 430px;
```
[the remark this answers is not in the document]
pfista: Yes it should scale the viewport down below that size. Worth testing on the actual device to be sure though!

apps/daimo-web/src/components/PixelBackground.tsx
pfista: I just pulled this out to be a static asset, though it only decreased first load js bundle size by 6kb

PR #756, opened 2024-02-25
mobile: make contact search unique, fix list key bug

nalinbhardwaj:
Tested on iOS simulator with:
- Two system contacts with same name, different email -> both show up and clickthrough correctly
- Two system contacts with same name, same email -> one shows up and click through correctly
- Two system contacts with different name, same email -> only one shows up and click through correctly

### DIFF
--- a/apps/daimo-mobile/src/logic/daimoContacts.ts
+++ b/apps/daimo-mobile/src/logic/daimoContacts.ts
@@ -45,6 +45,20 @@ export type DaimoContact = EAccountContact | EmailContact | PhoneNumberContact;
 // on-chain account)
 export type MsgContact = EmailContact | PhoneNumberContact;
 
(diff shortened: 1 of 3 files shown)


apps/daimo-mobile/src/logic/daimoContacts.ts
```diff
@@ -45,6 +45,20 @@ export type DaimoContact = EAccountContact | EmailContact | PhoneNumberContact;
...
+export function getDaimoContactKey(contact: DaimoContact): string {
+  switch (contact.type) {
+    case "eAcc":
+      return contact.addr;
+    case "email":
+      return contact.email;
+    case "phoneNumber":
+      return contact.phoneNumber;
+  }
+}
```
dcposch, line 60: 🚀

dcposch (approved):
Looks great

PR #757, opened 2024-02-25
mobile email send: multiplex email clients

### LINKED ISSUE #644: Compose improvements
- iOS with Mail removed: "Reinstall Mail" prompt
- Non-Pixel Androids: showing share sheet instead of prefilled email/sms

### LINKED ISSUE #754: Copy to clipboard escape hatch
From Liam:

> I don’t have the mail app but I want to send the link. There doesn’t seem to be any way to copy the link to clipboard or anything

At the end of the Send to Contact flow:

image

nalinbhardwaj:
Closes #754 
Closes #644 

Tested on real iOS and Android device, but didn't record video.

### DIFF
--- a/apps/daimo-mobile/src/view/shared/composeSend.tsx
+++ b/apps/daimo-mobile/src/view/shared/composeSend.tsx
@@ -1,5 +1,6 @@
 import { EmailAddress, PhoneNumber } from "@daimo/common";
 import { Linking, Platform } from "react-native";
+import { getEmailClients, openComposer } from "react-native-email-link";
 
 type PaymentLinkComposeParams = {
   type: "paymentLink";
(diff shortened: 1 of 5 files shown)


apps/daimo-mobile/src/view/shared/composeSend.tsx
```diff
@@ -34,20 +30,20 @@ export async function composeEmail(
...
+    const appId = emailClients.length === 1 ? emailClients[0].id : undefined;
 
-    if (await Linking.canOpenURL(openString)) {
-      await Linking.openURL(openString);
-      return true;
-    } else {
-      return false;
-    }
+    const appUsed = await openComposer({
+      defaultEmailLabel: "Apple Mail", // Only used on iOS
```
dcposch, line 37: you're sure this is always apple mail?

dcposch (approved):
LGTM

apps/daimo-mobile/src/view/shared/composeSend.tsx
nalinbhardwaj: not sure for the edge case where someone has a different default email app (not equal to Apple Mail):

In that case, we'll display the ActionSheet anyway, and worst case clicking this button does the same thing as another button.

PR #763, opened 2024-02-26
api: fix Farcaster link not showing

nalinbhardwaj:
Tested on iOS simulator, looks like we weren't attaching the linkedAccounts to `EAccount`s from some API requests before -- this is nice because it gives us a negative diff also


dcposch (approved):
LGreatTM

Issue #665, opened 2024-01-30
Nav frozen


dcposch:
## Summary

**Occasionally, we observe the app getting into a state where nothing responds to tap or swipe.**

No debug logs yet.

See #538 for a possibly related, prior bug.

dcposch:
See https://github.com/software-mansion/react-native-reanimated/issues/3331
and https://github.com/daimo-eth/daimo/issues/538#issuecomment-1864036144

dcposch:
## Steps to reproduce

@nalinbhardwaj found reliable repro

* Shake phone to bring up the Send Debug Log bottom sheet
* Immediately, while sheet is opening, tap outside the sheet

Expected behavior: sheet either open or closed

Actual behavior: sheet disappears, app is now frozen and non-interactive

dcposch:
## We still have other instances where nav freezes

@nalinbhardwaj is able to reproduce the following bug about a quarter of the time:

* Close app entirely
* Tap on a request deep link

Expected behavior: app opens, goes to send tab.

Actual (intermittent) behavior: app opens, starts going to send tab, gets stuck:

![image](https://github.com/daimo-eth/daimo/assets/169280/a41eccff-67b0-4329-b015-0c0f1cf03b33)

dcposch:
Mitigated in #775

PR #775, opened 2024-02-27
mobile: fix freezing material navigator by using standard bottom tab

nalinbhardwaj:
another attempt at #665 -- tested on iOS and android simulator looks fine


dcposch (approved):
LGTM 🫡

PR #781, opened 2024-02-27
Improve hero phone appearance while animation loads

pfista:
This adds a white background within the phone frame in the hero section of the website to improve the appearance during page load (so the user doesn't see a phone frame without the center portion before the animation loads).

On a separate note: the gif is large, so we should consider optimizing it or swapping for a more efficient file format to improve the loading time and quality of the animation.


dcposch (approved):
👍

PR #787, opened 2024-02-29
alerts: faucet USDC balance

nalinbhardwaj:
Prevent future API downtime

### DIFF
--- a/packages/daimo-api/src/server/cron.ts
+++ b/packages/daimo-api/src/server/cron.ts
@@ -7,8 +7,9 @@ import {
   daimoChainFromId,
   daimoPaymasterV2Address,
   entryPointABI,
+  daimoPaymasterV2ABI,
+  erc20ABI,
 } from "@daimo/contract";
-import { daimoPaymasterV2ABI } from "@daimo/contract/dist/generated";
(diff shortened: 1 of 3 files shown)


dcposch (approved):
LGTM!

packages/daimo-api/src/server/cron.ts
```diff
@@ -111,12 +112,30 @@ export class Crontab {
...
+      args: [faucetAddr],
+    });
+    const balanceDollars = Number(amountToDollars(balanceUSDC));
+    console.log(`[CRON] checked faucet USDC balance ${balanceDollars}`);
+
+    await this.sendLowBalanceMessage(
+      balanceDollars,
+      `Faucet ${faucetAddr} USDC`,
+      250,
+      25
```
dcposch, line 137: recommend
```
isTestnet ? 20 : 250,
isTestnet ? 5 : 25
```
... since we are low on base sepolia usdc

nalinbhardwaj: hm I think it's kind of good to get the cron / notifs to match closely since those alerts go into a separate slack channel anyway. will leave as is unless you feel strongly.

PR #789, opened 2024-02-29
api: Upgrade Available CTA

dcposch:
image

### DIFF
--- a/packages/daimo-api/src/server/appVersion.ts
+++ b/packages/daimo-api/src/server/appVersion.ts
@@ -1,0 +1,49 @@
+import { assertEqual } from "@daimo/common";
+
+let appVersionTracker = null as AppVersionTracker | null;
+
+export function getAppVersionTracker() {
+  if (appVersionTracker == null) {
+    appVersionTracker = new AppVersionTracker();
+  }
+  return appVersionTracker;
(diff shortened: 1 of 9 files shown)


packages/daimo-api/src/server/appVersion.ts
```diff
@@ -0,0 +1,49 @@
...
+      console.log(
+        `[APP-VERSION] got latest iOS version: ${this.latestVersion}`
+      );
+    } catch (e) {
+      console.error(`[APP-VERSION] error checking iOS version`, e);
+    }
+  }
+
+  async getLatestVersionIOS() {
+    const url = "https://itunes.apple.com/lookup?bundleId=com.daimo";
```
nalinbhardwaj, line 39: maybe we should read the android version separately?
Once we submit app to review it could take variable time to land on Android vs iOS and it might prompt banner for users without an actual update.

dcposch: will do in a followup.. getting the latest android version is harder.

dcposch:
deployed to prod so that people update to latest app version.

cc @nalinbhardwaj for post review

nalinbhardwaj (approved):

PR #790, opened 2024-03-01
api: only show tg suggestion if latest sync has transfer from someone…

nalinbhardwaj:
… else

tested on iOS simulator


nalinbhardwaj:
tested on iOS simulator, will push to prod, PTAL post-deploy @dcposch

PR #800, opened 2024-03-04
new download page

nalinbhardwaj:
new flow:
- `/download` instead of `/l/download`: no deep linking into the app
- on non iOS / android devices, redirect to `/download/other` -- show all three possibilities

Similar to signal.org.

TODO:
- [x] Pick an icon to represent both categories that's not bad, cc @michelle-chiu want to send me one and we can merge this

demo on various screen size: https://streamable.com/ip7vgj


michelle-chiu:
@nalinbhardwaj Looks great :)

The only ask is if you could change the iPhone/iPad icon to this one, so the stroke weight matches the one on desktop? 

smartphone

nalinbhardwaj:
Changed icon to be from pack after discussing with @michelle-chiu IRL

apps/daimo-web/src/components/buttons.tsx
```diff
@@ -29,17 +29,57 @@ export function PrimaryOpenInAppButton({
...
+    case "danger":
+      return "text-danger border-danger";
+    case "success":
+      return "text-success border-success";
+    default:
+      return "text-primaryLight border-primaryLight";
+  }
+}
+
+export function PrimaryButton({
```
dcposch, line 62: 👍

dcposch (approved):
LGTM

PR #802, opened 2024-03-06
mobile: include recent senders in search, improve /l/invite 

dcposch:
* show inviter when opening invite deeplink
  * currently you just stay on the home screen, which feels like a bug
* include recent senders in search

### Recent senders show up

Before this PR, this list would've been empty. (No recent recipients, 1 recent sender, now shown:)

image

### "First time paying" warning is still there

If someone has paid you but you've never paid them before, you see the following:

Screenshot 2024-03-05 at 18 49 17


apps/daimo-mobile/src/view/screen/AccountScreen.tsx
```diff
@@ -1,6 +1,9 @@
+import { getTeamDaimoFaucetAcc } from "@daimo/api/src/contract/nameRegistry";
```
nalinbhardwaj, line 1: want to move to @daimo/common or elsewhere?

nalinbhardwaj (approved):

PR #814, opened 2024-03-07
Add FC profile images

koredefashokun:
This PR displays Farcaster profile images instead of standard contact bubbles, where available.

| Home | Search | Send |
|----------|-----------|------------|
| simulator_screenshot_BFE60C19-3A76-4246-BCB4-A878EA769435 | simulator_screenshot_9F2D20CD-2335-45BE-8916-FF14FBDC1E5C | simulator_screenshot_D3E919C9-4B0E-4F46-8B39-F298B1A742DD|


apps/daimo-mobile/package.json
```diff
@@ -97,7 +97,8 @@
     "text-encoding-polyfill": "^0.6.7",
     "ts-unused-exports": "^9.0.4",
     "typescript": "5.2.2",
-    "viem": "^1.19.15"
+    "viem": "^1.19.15",
+    "expo-image": "~1.3.5"
```
dcposch, line 101: 👍

apps/daimo-mobile/src/view/shared/ContactBubble.tsx
```diff
@@ -103,11 +114,25 @@ export function Bubble({
...
+  );
+
   return (
-    <View style={style}>
-      <Text style={textStyle} numberOfLines={1} allowFontScaling={false}>
-        {children}
-      </Text>
+    <View style={{ width: size, height: size }}>
+      <Image source={image} style={imageStyle} />
+      <View style={style}>
```
dcposch, line 131: do we want both the image and the letter? i thought just one or the other

apps/daimo-web/package.json
```diff
@@ -28,6 +28,7 @@
     "pino-pretty": "^10.2.3",
     "react": "18.2.0",
     "react-dom": "^18.2.0",
+    "sharp": "^0.33.2",
```
dcposch, line 31: 👍 fixes console warnings?

apps/daimo-mobile/src/view/shared/SearchHeader.tsx
```diff
@@ -86,8 +86,7 @@ export function SearchHeader({
   if (account == null) return null;
   const eAcc: EAccountContact = {
     type: "eAcc",
-    addr: account.address,
-    name: account.name,
+    ...toEAccount(account),
```
dcposch, line 89: could be simplified to `const eAcc: EAccountContact = toEAccount(account)` ?

apps/daimo-web/src/app/profile/[addr]/route.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+      const result = await sharp(buffer).resize(64, 64).toBuffer();
+      const resultBlob = new Blob([result], { type: "image/jpeg" });
+      const response = new NextResponse(resultBlob);
+      response.headers.set("Content-Type", "image/jpeg");
+
+      return response;
+    }
+
+    return NextResponse.json({ error: "No image found" }, { status: 400 });
```
Comment (dcposch) [inline apps/daimo-web/src/app/profile/[addr]/route.tsx:29]: nit: recommend returning 404 here

apps/daimo-web/src/app/profile/[addr]/route.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    if (profilePicture) {
+      const data = await fetch(profilePicture);
+      const blob = await data.blob();
+      const arrayBuffer = await blob.arrayBuffer();
+      const buffer = Buffer.from(arrayBuffer);
+
+      const result = await sharp(buffer).resize(64, 64).toBuffer();
+      const resultBlob = new Blob([result], { type: "image/jpeg" });
+      const response = new NextResponse(resultBlob);
```
Comment (dcposch) [inline apps/daimo-web/src/app/profile/[addr]/route.tsx:23]: looks great

we should cache (addr > Blob | null) , either in memory lru cache or db table

& either here, or in a followup pr

packages/daimo-api/src/api/profile.ts
```diff
@@ -134,6 +135,10 @@ export class ProfileCache {
     return this.linkedAccounts.get(addr) || [];
   }
 
+  getProfilePicture(addr: Address) {
+    return `${daimoDomainAddress}/profile/${addr}`;
```
dcposch, line 139: nit but i recommend making this `${daimoDomainAddress}/profile/${addr}/pfp.png`

...so that we can put a more complete profile at `/profile/<addr>` in future

dcposch (commented):
Looks great to me

Please post a screenshot from simulator

apps/daimo-mobile/src/view/shared/ContactBubble.tsx
koredefashokun: In this case, the letter is the fallback. If we make it one or the other, a broken image (e.g. no PFP) means that we get a blank view. Using `<Image>`'s `onError` prop to set a "failed image fetch" state also doesn't work, since it waits until subsequent renders to update to the text.

packages/daimo-api/src/api/profile.ts
koredefashokun: Good point. I'll make the change.

apps/daimo-web/src/app/profile/[addr]/route.tsx
Comment (koredefashokun) [reply apps/daimo-web/src/app/profile/[addr]/route.tsx:23]: Next.js apparently caches the results of `fetch`-es like this one. But I agree that using a LRU cache is more explicit and gives more control over what is cached, where and for how long.

apps/daimo-mobile/src/view/shared/SearchHeader.tsx
koredefashokun: `toEAccount` returns an `EAccount`, not an `EAccountContact`. I'll make a `toEAccountContact` function

apps/daimo-web/package.json
koredefashokun: Actually, it's for the image resizing. I tried to get imagescript to work, but my efforts were abortive.

apps/daimo-web/src/app/profile/[addr]/route.tsx
Comment (koredefashokun) [reply apps/daimo-web/src/app/profile/[addr]/route.tsx]: Fair point. Updated.

packages/daimo-api/src/api/profile.ts
```diff
@@ -134,6 +135,10 @@ export class ProfileCache {
     return this.linkedAccounts.get(addr) || [];
   }
 
+  getProfilePicture(addr: Address) {
+    return `${daimoDomainAddress}/profile/${addr}/pfp.png`;
```
nalinbhardwaj, line 139: hm I see `{ type: "image/jpeg" }` above, is this a typo?

koredefashokun: Yes, it is. Thanks. Fixed in the latest commit

PR #820, opened 2024-03-08
Update expo 49 -> 50

wojtus7:
After an uphill struggle there it is building and alive 

(I needed ot remove node modules and rebuild projects with --clear to make it work)

Android: 
image

iOS:
image


apps/daimo-mobile/package.json
```diff
@@ -51,52 +51,52 @@
...
-    "react-native-safe-area-context": "4.6.3",
-    "react-native-screens": "^3.27.0",
-    "react-native-shake": "^5.5.2",
-    "react-native-svg": "13.9.0",
+    "react-native-reanimated": "~3.6.2",
+    "react-native-safe-area-context": "4.8.2",
+    "react-native-screens": "~3.29.0",
+    "react-native-shake": "^5.6.0",
+    "react-native-svg": "14.1.0",
     "react-native-url-polyfill": "^2.0.0",
```
nalinbhardwaj, line 95: We might not need this anymore? Acc to the blog https://expo.dev/changelog/2024/01-18-sdk-50

package.json
```diff
@@ -26,6 +26,10 @@
     "packages/*",
     "apps/*"
   ],
+  "overrides": {
```
nalinbhardwaj, line 29: is this necessary?

nalinbhardwaj (changes requested):
General PR looks great, want to fix CI complaints?

package.json
wojtus7: React-native-shake use react-native 72 instead of 73 as a devdependency and it would break whole app during build so I enforce it to use 0.73

It looks like it works fine but if we want to remove it we need to remove the library, and the shake feature or find some different library to detect the gesture

apps/daimo-mobile/babel.config.js
```diff
@@ -2,7 +2,7 @@
 module.exports = function (api) {
   api.cache(true);
   return {
-    presets: ["babel-preset-expo", "module:metro-react-native-babel-preset"],
+    presets: ["babel-preset-expo"],
```
dcposch, line 5: 👍

apps/daimo-mobile/package.json
dcposch: ya can try deleting ^

dcposch (approved):
LGreatTM

PR #845, opened 2024-03-13
payment link: make history op cancel work, make back not cross

nalinbhardwaj:
checked in sim

https://streamable.com/70i42d

### DIFF
--- a/apps/daimo-mobile/src/view/screen/link/NoteScreen.tsx
+++ b/apps/daimo-mobile/src/view/screen/link/NoteScreen.tsx
@@ -25,7 +25,7 @@ import {
   transferAccountTransform,
   useSendAsync,
 } from "../../../action/useSendAsync";
-import { ParamListHome, useExitToHome } from "../../../common/nav";
+import { ParamListHome, useExitBack } from "../../../common/nav";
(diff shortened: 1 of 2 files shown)


apps/daimo-mobile/src/view/screen/link/NoteScreen.tsx
```diff
@@ -216,10 +221,11 @@ function NoteDisplayInner({
   const netDollarsReceivedStr = getAmountText({ dollars: netRecv });
 
   // On success, go home, show newly created transaction
-  const goHome = useExitToHome();
+  const backHome = useExitBack();
```
dcposch, line 224: `goBack`? this won't necessarily go all the way to the home screen.

dcposch (approved):
Otherwise LGTM

apps/daimo-mobile/src/view/screen/link/NoteScreen.tsx
nalinbhardwaj: Since `NoteScreen` is in HomeTab now I believe it'll always go to Home Screen, but edited name nonetheless.

PR #861, opened 2024-03-15
Don't close search when accidentally dismiss keyboard

### LINKED ISSUE #829: Scrolling on home screen resets spuriously
https://github.com/daimo-eth/daimo/assets/6984346/b831fcc5-69a5-4fdb-bd14-70dae4762c47

wojtus7:
Closes #829

The reason why thing in the video happens is that sometimes it's possible to accidentally close the keyboard and it results in closing whole search list.

This PR makes it so it only do so if the input is empty or the back arrow is pressed.
On refocus it clears the input right now but it can be easily changed.

Also the keybaord padding can be animated using useAnimatedKeyboard but wanted to let this PR be simpler and change it in future PR if necessary

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/107ca88b-2d11-4a04-b04b-0e25f3da7d94

### DIFF
--- a/apps/daimo-mobile/src/view/shared/SearchHeader.tsx
+++ b/apps/daimo-mobile/src/view/shared/SearchHeader.tsx
@@ -1,5 +1,5 @@
 import Octicons from "@expo/vector-icons/Octicons";
-import { RefObject, useCallback, useEffect } from "react";
+import { RefObject, useCallback, useEffect, useRef } from "react";
 import { Keyboard, StyleSheet, TextInput, View } from "react-native";
 import { TouchableOpacity } from "react-native-gesture-handler";
 import Animated, {
@@ -8,7 +8,10 @@ import Animated, {
   withTiming,
 } from "react-native-reanimated";
 
-import { AnimatedSearchInput } from "./AnimatedSearchInput";
+import {
+  AnimatedSearchInput,
+  AnimatedSearchInputRef,
+} from "./AnimatedSearchInput";
 import { ButtonCircle } from "./ButtonCircle";
 import { ContactBubble } from "./ContactBubble";
 import { color } from "./style";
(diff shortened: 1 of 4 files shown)


apps/daimo-mobile/src/view/shared/SearchHeader.tsx
```diff
@@ -89,7 +93,14 @@ export function SearchHeader({
   return (
     <View style={styles.header}>
       <Animated.View key="back" style={backButton}>
-        <TouchableOpacity onPress={() => Keyboard.dismiss()} hitSlop={16}>
+        <TouchableOpacity
+          onPress={() => {
+            setPrefix(undefined);
+            Keyboard.dismiss();
+            ref.current?.closeInput();
```
dcposch, line 100: hmm im curious how this was working before (cancel keyboard > search prefix remains in place?)

apps/daimo-mobile/src/view/shared/SearchHeader.tsx
```diff
@@ -99,12 +110,13 @@ export function SearchHeader({
...
       </Animated.View>
       <AnimatedSearchInput
+        ref={ref}
         icon="search"
         placeholder="Search for user..."
         value={prefix || ""}
         onChange={setPrefix}
-        onFocus={() => setPrefix("")}
-        onBlur={() => setPrefix(undefined)}
+        onFocus={() => setPrefix(prefix || "")}
```
dcposch, line 118: might not need to clear on focus anymore (now that we're clearing prefix when the back button is pressed)

dcposch (approved):
One small comment--otherwise LGTM--ship it!

apps/daimo-mobile/src/view/shared/SearchHeader.tsx
wojtus7: Blur keyboard always set the prefix to `undefined`
Focus set to `""`
And then typing until keyboard dismiss of keyboard then `undefined` right again

PR #864, opened 2024-03-16
bot: fix tsconfig, drop expo prepare

nalinbhardwaj:
- fixes tsconfig errors in bot that caused `npm run start` to be failing on fresh install
- removes `expo-module prepare` from our expo modules.
    - It seems that's using yarn under the hood instead of npm correctly, causing build failures. We don't need it ATM given that we don't publish our expo modules to NPM + from what I can tell it only runs misc. nice to haves rather than anything mission critical, see https://github.com/expo/expo/blob/main/packages/expo-module-scripts/README.md#setup

Tested Render build no longer fails when pointed at this branch: https://daimo-farcaster-bot.onrender.com


apps/daimo-bot/tsconfig.json
```diff
@@ -1,110 +1,7 @@
...
-    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
-
-    /* Completeness */
-    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
-    "skipLibCheck": true /* Skip type checking all .d.ts files. */
-  },
-  "exclude": ["**/test/*.ts"]
+    "strict": true,
+    "resolveJsonModule": true
+  }
```
dcposch, line 6: i suggested this on on the original PR! cc @panieldark

dcposch (approved):
LGreatTM, nice cleanup

PR #867, opened 2024-03-18
web: invite frame

dcposch:


dcposch:
## Farcon invite 

image

## Farcaster pre-permissionless invite

TO DO
- [ ] Nice cover. Make it very obvious what it does, no mystery.
- [ ] Success screen asks for recast.

Motivating people to post screenshots after it works would also be great.
Maybe we post a end-to-end onboarding video ourselves.

nalinbhardwaj (approved):

apps/daimo-bot/.env.example
```diff
@@ -1,5 +1,6 @@
-NEYNAR_API_KEY="<replaceme uuid>"
+DAIMO_NEYNAR_KEY="<replaceme uuid>"
```
nalinbhardwaj, line 1: probably need to update render environments

PR #889, opened 2024-03-20
trpc: correct context

nalinbhardwaj:
Went very deep into the TRPC codebase and it turns out they do actually still need this -- only tanstack upstream has removed it, they haven't.


nalinbhardwaj:
Merged into #890

dcposch:
superseded by #890

Issue #884, opened 2024-03-20
Internal canary android Home Screen is broken?


nalinbhardwaj:
https://github.com/daimo-eth/daimo/assets/6984346/f9b381e0-0ede-4cd6-989e-12f0c3b752a6

nalinbhardwaj:
Fixed in #894

PR #894, opened 2024-03-21
Don't close search when accidentally dismiss keyboard (and don't break android along the way)

### LINKED ISSUE #829: Scrolling on home screen resets spuriously

wojtus7:
Close #829 together with #884

Sorry about android bug, I missused styles in wrong component 🫠

Android:

https://github.com/daimo-eth/daimo/assets/42337257/46fc493d-17fd-4856-accf-dfba3a66a554

iOS:

https://github.com/daimo-eth/daimo/assets/42337257/7a33bf02-3b40-4da4-96ec-043f59e0ff28

### DIFF
--- a/apps/daimo-mobile/src/view/shared/AnimatedSearchInput.tsx
+++ b/apps/daimo-mobile/src/view/shared/AnimatedSearchInput.tsx
@@ -1,12 +1,5 @@
 import Octicons from "@expo/vector-icons/Octicons";
-import {
-  RefObject,
-  forwardRef,
-  useCallback,
-  useImperativeHandle,
-  useRef,
-  useState,
-} from "react";
+import { RefObject, useCallback, useEffect, useRef, useState } from "react";
 import {
   Dimensions,
   LayoutChangeEvent,
@@ -36,7 +29,7 @@ const BACK_ICON = 30 + 8;
 const animationConfig = { duration: 150 };
 
 interface AnimatedSearchInputProps {
-  value: string;
+  value?: string;
   onChange: (value: string) => void;
   onFocus?: () => void;
   onBlur?: () => void;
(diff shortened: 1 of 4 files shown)


apps/daimo-mobile/src/view/shared/AnimatedSearchInput.tsx
```diff
@@ -28,126 +35,170 @@ const BACK_ICON = 30 + 8;
...
+          openInput() {
+            animateOpenInput();
+          },
+          closeInput() {
+            animateCloseInput();
+          },
+        };
+      },
+      []
+    );
```
dcposch, line 133: what does useImperativeHandle do?

/ where does `openInput()` actually get called?

apps/daimo-mobile/src/view/shared/AnimatedSearchInput.tsx
```diff
@@ -28,126 +35,170 @@ const BACK_ICON = 30 + 8;
...
-    </View>
-  );
 }
 
+export type AnimatedSearchInputRef = {
+  openInput: () => void;
+  closeInput: () => void;
+};
+
+export const AnimatedSearchInput = forwardRef<
```
dcposch, line 57: forwardRef strikes me as complicated, but OK if there's no better way

apps/daimo-mobile/src/view/shared/SearchHeader.tsx
```diff
@@ -89,7 +93,14 @@ export function SearchHeader({
   return (
     <View style={styles.header}>
       <Animated.View key="back" style={backButton}>
-        <TouchableOpacity onPress={() => Keyboard.dismiss()} hitSlop={16}>
+        <TouchableOpacity
+          onPress={() => {
+            setPrefix(undefined);
+            Keyboard.dismiss();
+            ref.current?.closeInput();
```
dcposch, line 100: could we avoid the whole `forwardRef` with something like

```ts
useEffect(() => {
  if (prefix === undefined) animateCloseInput();
}, [prefix]);
```

..insider AnimatedSearchInput?

dcposch (commented):
This is just recreating #861 but with an Android fix, right?

Video looks good to me.

Recommend avoiding the forwardRef if possible, since i think that might save us a lot of code---see comment

apps/daimo-mobile/src/view/shared/SearchHeader.tsx
wojtus7: I though the problem here was that it is parsed as `value={prefix || ""}` because it needs to always use a string but I think we can change that and you are right that we can remove forwardRef because of that.

Thanks for pointing that out 🫡

dcposch (approved):
Nice cleanup. Looks great!

PR #912, opened 2024-03-25
telemetry: log reverted txes, increase paymaster warns

### LINKED ISSUE #891: Observability: link directly to correct debug trace on every failed or reverted op
## Goal

Whenever a user tries to transact and fails or the transaction reverts, we should get:
- exactly one alert, and
- the alert should link directly to a correct and complete trace, such as [this one](https://basescan.org/vmtrace?txhash=0x07a55b03e90b0c8fe555b2fb3ce76680ba5c66f7966b17ca0301a28643c2b022&type=gethtrace2)

nalinbhardwaj:
Closes #891 

image
^purposefully broke signatures sent by client

### DIFF
--- a/packages/daimo-api/src/env.ts
+++ b/packages/daimo-api/src/env.ts
@@ -1,5 +1,10 @@
 import { DaimoChain, getChainConfig } from "@daimo/contract";
 
+// Enable JSON.stringify for BigInts
+(BigInt.prototype as any).toJSON = function () {
+  return this.toString();
+};
+
 export const chainConfig = getChainConfig(
   (process.env.NEXT_PUBLIC_DAIMO_CHAIN || "baseSepolia") as DaimoChain
 );
--- a/packages/daimo-api/src/network/viemClient.ts
+++ b/packages/daimo-api/src/network/viemClient.ts
@@ -19,13 +19,14 @@ import {
 import { privateKeyToAccount } from "viem/accounts";
 
 import { chainConfig } from "../env";
+import { Telemetry } from "../server/telemetry";
 import { memoize } from "../utils/func";
 
 /**
  * Loads a wallet from the local DAIMO_API_PRIVATE_KEY env var.
  * This account sponsors gas for account creation (and a faucet, on testnet).
  */
-export function getViemClientFromEnv() {
+export function getViemClientFromEnv(monitor: Telemetry) {
   // Connect to L1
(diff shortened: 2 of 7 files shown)


packages/daimo-api/src/env.ts
```diff
@@ -1,5 +1,10 @@
 import { DaimoChain, getChainConfig } from "@daimo/contract";
 
+// Enable JSON.stringify for BigInts
```
dcposch, line 3: i'm very dubious of these kinds of global edits, but in this case i think it's OK

packages/daimo-api/src/network/viemClient.ts
```diff
@@ -83,6 +85,30 @@ export class ViemClient {
...
+    try {
+      const receipt = await this.publicClient.waitForTransactionReceipt({
+        hash,
+      });
+      console.log(`[CHAIN] waitForReceipt ${hash}: ${JSON.stringify(receipt)}`);
+      if (receipt.status !== "success") {
+        this.onReceiptError(hash, JSON.stringify(receipt));
+      }
+    } catch (e) {
+      console.log(`[CHAIN] waitForReceipt error: ${e}`);
```
dcposch, line 107: console.error

packages/daimo-api/src/server/cron.ts
```diff
@@ -111,8 +111,8 @@ export class Crontab {
     await this.sendLowBalanceMessage(
       depositEth,
       `Paymaster ${daimoPaymasterV2Address} ETH`,
-      isMetaPaymasterEnabled ? 0.01 : 0.03,
-      isMetaPaymasterEnabled ? 0.005 : 0.01
+      isMetaPaymasterEnabled ? 0.01 : 0.1,
+      isMetaPaymasterEnabled ? 0.005 : 0.2
```
dcposch, line 115: 👍

packages/daimo-api/src/env.ts
nalinbhardwaj: Agree - I think in this case there's a great chance of bugs being caused by console logs/warnings/errors crashing due to lack of this rather than the other way around.

packages/daimo-api/src/network/viemClient.ts
nalinbhardwaj: done

PR #919, opened 2024-03-27
api: invite frame: one per fid

dcposch:
## Summary

Prevent griefing by users going thru the invite frame many times, then bypassing one-per-device.


nalinbhardwaj (approved):
lgtm minus todos

dcposch:
## Club Ted invite frame

image

PR #936, opened 2024-03-31
viem client: save eth_chainId roundtrip

### LINKED ISSUE #935: Upstream viem bug: tx chain ID != client chain ID
seems to be due to https://github.com/wevm/viem/issues/1389 -- happened once so far. Either fix here or upstream PR.

nalinbhardwaj:
Tested on local API saves ~50ms.

Likely fixes #935 as well -- though the exact cause of that is still unclear to me.


dcposch (approved):
🔥

Issue #930, opened 2024-03-28
add to F-Droid


antonok-edm:
I know it's early, but it'd be great to eventually see Daimo available on [F-Droid](https://f-droid.org/en/) for folks who aren't using Google Play Services on their Android phones.

antonok-edm:
Releasing on F-Droid will require building on Linux without EAS.

`expo` doesn't appear to support local builds at all, from what I could tell - `expo build:android` gives a message about how `eas` should be used instead. I had to use `npx expo prebuild` followed by `react-native build-android` directly.

ottok:
If you can't publish on F-Droid, can you alternatively make a plain .apk available for download?

Many tech pioneers use innovative non-standard platforms and tools that may still be in their infancy. By publishing on F-Droid or making a plain .apk available, you could reach all those audiences that run e.g. Dalvik or Sailfish or PinePhone or Purism etc.

PR #960, opened 2024-04-03
`npm install` improvements

antonok-edm:
`npm install` doesn't automatically run `build` scripts, but it [does](https://docs.npmjs.com/cli/v9/using-npm/scripts#npm-install) run `prepare` scripts.

Not sure how you were previously handling this (maybe `eas` has some special handling?), but this change should compile all of the `@daimo/*` packages when running `npm install` in the repo.

I've also updated `@daimo/api` to use the same tsconfig as the other packages (uncovering a type bug in the process!).

### DIFF
--- a/apps/daimo-mobile/tsconfig.json
+++ b/apps/daimo-mobile/tsconfig.json
@@ -3,5 +3,14 @@
   "compilerOptions": {
     "strict": true
   },
-  "include": ["./*.js", "./src/**/*", "./test/**/*", "app.config.ts"]
+  "include": ["./*.js", "./src/**/*", "./test/**/*", "app.config.ts"],
+  "references": [
+    { "path": "../../packages/daimo-api" },
+    { "path": "../../packages/daimo-common" },
+    { "path": "../../packages/daimo-userop" },
+    { "path": "../../packages/daimo-contract" },
+    { "path": "../../packages/daimo-expo-enclave" },
+    { "path": "../../packages/daimo-expo-passkeys" },
+    { "path": "../../packages/daimo-expo-app-delegate" }
+  ]
 }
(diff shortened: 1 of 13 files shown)


dcposch (approved):
Looks great, thanks!

antonok-edm:
> Not sure why this caught an extra type issue when `npm test` runs `tsc` already

`validateMemo` is exported, but it returns `MemoStatus`, which wasn't previously exported as a type.

Could be because of `--transpile-only`? I'm not super familiar with how that works.

> I don't think we need the "prepare" script. CI is unhappy now, probably because of that

Super weird, `npm ci` fails on a fresh checkout of this branch, but then running it a second time succeeds :shrug:

I'll try to look into this further later.

dcposch:
Common pattern I've seen with npm is

```
npm i # or npm ci
npm run build
```

& we're doing that before invoking eas, so I think that part's fine?

antonok-edm:
> so I think that part's fine?

What I'm trying to address is that when I'm following [these steps](https://github.com/daimo-eth/daimo/issues/930#issuecomment-2035832619) to build locally and launching the app, there's immediately a runtime error because it tries to import from `dist` folders that don't exist yet (i.e. `tsc` hasn't been run in any of the dependencies).

I did find a [couple](https://earthly.dev/blog/setup-typescript-monorepo/) of [references](https://spin.atomicobject.com/typescript-project-workspaces/) to using the `composite` setting for `tsconfig.json`, which at least builds all of the dependencies recursively with a single `tsc --build` invokation from `apps/daimo-mobile`. Just updated here. Can we give that a try on CI to make sure it doesn't mess with anything else?

dcposch:
Can revisit

PR #973, opened 2024-04-05
Dynamic frame image for invites

### LINKED ISSUE #971: Improve Farcaster invite frame
## Goal

We're about to ship a `SHARE FRAME ON FARCASTER` button on the invite tab.
It works, but it's not as good as it could be.

## Current functionality

Frame URL format: `https://daimo.com/frame/invite/<inviter daimo address>`, eg
https://daimo.com/frame/invite/0x4D350d99364634e07B01a9986662787DD3755F0A

❌ Frame is a static image, so if Alice posts one frame (with her as the inviter) and Bob posts another, they look identical.
✅ Power Users get $10 on joining
🟡 Everyone else just gets a "You're not on the list" screen

image
image

## Proposed 

✅ Frame is personalized, using a generated image. See `preview/route.tsx` for an example of how to do this.
✅ Power Users get $10 on joining
✅ Everyone else still gets to join, just without a bonus.

Current text:
> ⚡️
> `POWER USER?`
> ### Claim your $10 USDC invite on Daimo

See image above.

Proposed
> ⚡️
> `+$10 BONUS FOR POWER USERS`
> ### Join nibnalin on Daimo

...where that last line is dynamic based on inviter eAcc.

Concretely, the frame callback will *always* return success and an invite link. It's just that:
- If the `interactor` on the frame click is a power user, then `bonusDollarsInvitee` is $10
- Otherwise, `bonusDollarsInvitee` is 0

koredefashokun:
This PR sets up a dynamic image for the Warpcast invite frame.

image

Closes #971


apps/daimo-web/src/app/frame/invite/[addr]/page.tsx
```diff
@@ -25,7 +25,9 @@ export async function generateMetadata(props: LinkProps): Promise<Metadata> {
   }
 
   // Show image, button text customized to this invite frame
-  const frameImg = getAbsoluteUrl("/assets/frame/InvInitPowerUser.png");
+  const frameImg = getAbsoluteUrl(
+    `/frame/invite/${eAcc.addr}/preview?name=${eAcc.name}`
```
Comment (dcposch) [inline apps/daimo-web/src/app/frame/invite/[addr]/page.tsx:29]: 🚀

apps/daimo-web/src/app/frame/invite/[addr]/preview/route.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const { searchParams } = new URL(request.url);
+
+  if (!searchParams.has("name")) {
+    throw new Error("Invalid preview params");
+  }
+  const name = searchParams.get("name");
+
+  return new ImageResponse(<InvitePreviewImg name={name!} />, {
+    width: 1086,
+    height: 569,
```
Comment (dcposch) [inline apps/daimo-web/src/app/frame/invite/[addr]/preview/route.tsx:17]: looks good, matches image dimensions

dcposch (approved):
🔥

PR #985, opened 2024-04-08
Add Sentry error reporting

koredefashokun:
This PR adds Sentry error reporting to the daimo api.

### DIFF
--- a/packages/daimo-api/src/server/router.ts
+++ b/packages/daimo-api/src/server/router.ts
@@ -12,6 +12,7 @@ import {
   zUserOpHex,
 } from "@daimo/common";
 import { SpanStatusCode } from "@opentelemetry/api";
+import * as Sentry from "@sentry/node";
 import { TRPCError } from "@trpc/server";
 import { getAddress, hexToNumber } from "viem";
 import { z } from "zod";
@@ -103,7 +104,12 @@ export function createRouter(
     return opts.next();
   });
 
+  const sentryMiddleware = trpcT.middleware(
+    Sentry.Handlers.trpcMiddleware({ attachRpcInput: true }) as any
+  );
+
   const publicProcedure = trpcT.procedure
+    .use(sentryMiddleware)
     .use(corsMiddleware)
     .use(tracerMiddleware)
     .use(readyMiddleware);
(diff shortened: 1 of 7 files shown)


packages/daimo-api/src/server/router.ts
```diff
@@ -103,7 +104,12 @@ export function createRouter(
     return opts.next();
   });
 
+  const sentryMiddleware = trpcT.middleware(
+    Sentry.Handlers.trpcMiddleware({ attachRpcInput: true }) as any
```
koredefashokun, line 108: The `any` type here is tied to this issue: https://github.com/getsentry/sentry-javascript/issues/9536

packages/daimo-api/src/server/telemetry.ts
```diff
@@ -56,6 +59,20 @@ export class Telemetry {
...
+      Sentry.init({
+        dsn: sentryDSN,
+        integrations: [
+          new Sentry.Integrations.Postgres(),
+          nodeProfilingIntegration(),
+        ],
+        tracesSampleRate: 1.0,
+        profilesSampleRate: 1.0,
+      });
+    }
```
koredefashokun, line 75: I chose to set up Sentry in the telemetry file for two reasons:
- Keeping the `server.ts` file clean and
- Ensuring that it remains amongst the first things that happen when the server starts

That said, I can move it to another file if it is considered too distinct from the general telemetry functions.

koredefashokun: There are also a number of interesting things that can be achieved with the Sentry config. It might be interesting to experiment with a couple of them in another PR (mostly around profiling and instrumentation)

koredefashokun:
The Socket security issues are because of Sentry's postinstall scripts and not any actual runtime problems.

packages/daimo-api/src/server/router.ts
```diff
@@ -103,7 +104,12 @@ export function createRouter(
...
+  );
+
   const publicProcedure = trpcT.procedure
+    .use(sentryMiddleware)
```
dcposch, line 112: 🚀

dcposch (approved):
Looks great to me

PR #994, opened 2024-04-12
Slack bot - Add Daimo invite management

koredefashokun:
This PR adds a webhook to handle four new slash commands on Slack that are geared towards managing Daimo invites.

The commands are:
- `/create-invite`: Creates a new Daimo invite with given arguments
- `/view-invite-status`: View current status of a given invite link
- `/set-max-uses`: Update the `maxUses` value of the invite link
- `/help`: List all commands and their input/ouput formats


dcposch (commented):
Looks great to me! Left suggestions

dcposch (approved):
🚀

PR #995, opened 2024-04-13
shovel up: update to 1.2, type everything properly

nalinbhardwaj:
Tested on testnet shovel, this is the current deployment there -- transfers and other indexers still appear to work correctly, diff between generated `config.json` and old `config.json` seems clean.

Moves config to `@daimo/contract` so will need prod env change on push.


packages/daimo-contract/package.json
```diff
@@ -5,15 +5,17 @@
...
     "build": "tsc"
   },
   "author": "",
   "license": "GPL-3.0-or-later",
   "devDependencies": {
+    "@indexsupply/shovel-config": "^0.0.9",
     "@pimlico/erc20-paymaster": "^0.1.1",
     "@tsconfig/node20": "^20.1.0",
     "@wagmi/cli": "^1.4.1",
+    "tsx": "^4.7.2",
```
dcposch, line 18: we're using ts-node for this in other spots, is tsx better?

packages/daimo-contract/src/shovel/config.json
```diff
@@ -0,0 +1,1110 @@
...
+          "name": "tx_hash",
+          "column": "tx_hash"
+        },
+        {
+          "name": "log_addr",
+          "column": "log_addr",
+          "filter_op": "contains",
+          "filter_arg": [
+            "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
+            "0x036CbD53842c5426634e7929541eC2318f3dCF7e"
```
dcposch, line 391: This is Base USDC + Base Sepolia USDC

If we're parameterizing by `$CHAIN_NAME` etc we might want to have a `$COIN_ADDR` too

packages/daimo-contract/src/shovel/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  namesIntegration,
+  keyAddedIntegration,
+  keyRemovedIntegration,
+  transfersIntegration,
+  noteCreatedIntegration,
+  noteRedeemedIntegration,
+  requestCreatedIntegration,
+  requestCancelledIntegration,
+  requestFulfilledIntegration,
+  userOpIntegration,
```
dcposch, line 34: beautiful

dcposch (approved):
LGreatTM

packages/daimo-contract/package.json
nalinbhardwaj: `ts-node` has some outstanding bugs in interop with packages built with `bun`, `tsx` seems to be the recommended way to fix them.

packages/daimo-contract/src/shovel/config.json
nalinbhardwaj: will leave for future

PR #1027, opened 2024-04-26
api: reduce ethIndexer load even more

nalinbhardwaj:
testing on prod 🖖


nalinbhardwaj:
Tested on prod / testnet render for a decent bit, merging so master doesn't have broken code. PTAL post-commit @dcposch

PR #1034, opened 2024-04-27
quickfix push notifier: don't send push notifs on dust

nalinbhardwaj:
tested on local sim with prod API

### DIFF
--- a/packages/daimo-api/src/contract/foreignCoinIndexer.ts
+++ b/packages/daimo-api/src/contract/foreignCoinIndexer.ts
@@ -1,4 +1,8 @@
-import { ProposedSwap, guessTimestampFromNum } from "@daimo/common";
+import {
+  ProposedSwap,
+  guessTimestampFromNum,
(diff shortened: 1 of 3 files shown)


packages/daimo-api/src/server/pushNotifier.ts
```diff
@@ -335,6 +337,8 @@ export class PushNotifier {
 
     const dollars = amountToDollars(swap[0].toAmount);
 
+    if (Number(dollars) < 0.5) return []; // Don't notify on dust, still shows in app.
```
dcposch, line 340: recommend `DUST_DOLLARS`, can set to 1 probably

dcposch (approved):
LGTM otherwise

packages/daimo-api/src/server/pushNotifier.ts
nalinbhardwaj: done

PR #1039, opened 2024-05-02
mobile: small UX improvements to deposit address flow

nalinbhardwaj:
- show ending letters of address on deposit screen: that's what users use to authenticate the address on copying it elsewhere, and its annoying right now to not be able to see those

image

- allow long press to copy raw address on qr screen -- noticed Daniel from Bountycaster do this and he explained he wanted this interaction since there's no easy way to get address from that screen.

Screenshot 2024-05-01 at 9 55 37 PM
Screenshot 2024-05-01 at 9 55 41 PM


dcposch (approved):
LGTM

PR #1040, opened 2024-05-02
uniswap client: workaround ethersJS bug

nalinbhardwaj:
Observed bug: foreign coin swaps not working at all, not for DOGINME and not for USDbC, unable to reproduce on running prod API server with simulator on my local.

on investigating logs:

```
[UNISWAP] fetching swap for 0x27785Ad361898B526F37d87C4fAcFD757Ff0622F 0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA
```
happens but no logs from `[UNISWAP]` even though we are supposed to always log returned information from the SDK -- it's also unclear if there was an error why `retryBackoff` did not trigger.

Googling finds other reports of similar issues, apparently from Ethers Websocket being broken.

Fix: use HTTP RPC for the Uniswap SDK instead, added env variables to prod and testnet APIs.

### DIFF
--- a/packages/daimo-api/src/network/uniswapClient.ts
+++ b/packages/daimo-api/src/network/uniswapClient.ts
@@ -62,11 +62,15 @@ export class UniswapClient {
   private swapCache: Map<string, ProposedSwap> = new Map();
 
   constructor() {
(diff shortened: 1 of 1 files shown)


nalinbhardwaj:
PTAL post-commit @dcposch

PR #1043, opened 2024-05-09
api: use balance from shovel instead of client

nalinbhardwaj:
Fixes:
image

Fixes balance out of sync of displayed transfers, unclear why that could happen.


dcposch (approved):
LGreatTM, nice simplification

PR #1050, opened 2024-05-15
mobile: use api.daimo.xyz, remove Render

dcposch:


dcposch:
Shipping, cc @nalinbhardwaj for post review

PR #1068, opened 2024-05-24
deploy api image via github actions

dehli:
Let me know if you'd like any changes. Builds and deploys api on each push to `master` and `prod`. Successful run can be seen [here](https://github.com/daimo-eth/daimo/actions/runs/9225477659/job/25383053568).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,10 +2,62 @@ name: CD
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
+      - prod
 
 jobs:
-  deploy:
+  api-deploy:
+    name: API Deploy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-region: ${{ secrets.AWS_REGION }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -2,10 +2,61 @@ name: CD
...
+        with:
+          platforms: arm64
+
+      - name: Build, tag, and push docker image to Amazon ECR
+        env:
+          GIT_SHA: ${{ github.sha }}
+          IMAGE_NAME: ${{ steps.login-ecr.outputs.registry }}/${{ vars.AWS_ECR_API_REPO }}
+        run: |
+          docker build --platform linux/arm64 -t $IMAGE_NAME:$GIT_SHA -t $IMAGE_NAME:$IMAGE_TAG -f packages/daimo-api/Dockerfile .
+          docker push $IMAGE_NAME --all-tags
```
dcposch, line 51: do we want `--all-tags` or just stage/prod tag, depending on branch?

dcposch (approved):
otherwise LGTM

.github/workflows/cd.yml
dehli: With the current code we tag the SHA in addition to latest/prod (depending on the branch) and push both tags up.

image

Would you prefer to just have the single latest/prod tag @dcposch?

dcposch: ah, i see.  that looks good

dcposch: as long as we're sure that a `master` build will never overwrite the `prod` tag

dehli: The action's container will always  run in a fresh state so there won't be more than those two tags however I went ahead and updated the command to be more explicit for clarity.

PR #1074, opened 2024-05-25
api: add usdt tron bridge

nalinbhardwaj:


apps/daimo-mobile/src/view/shared/Bubble.tsx
```diff
@@ -56,18 +56,14 @@ export function ContactBubble({
...
+        default:
           return (
             <Octicons
               name="arrow-switch"
               size={fontSize}
               color={color.primary}
             />
           );
-        default:
-          return "?";
```
nalinbhardwaj, line 70: ? is ugly, we should never show it imo

dcposch (approved):
Looks great

Ideally we'd have a bottomsheet to explain the min TRX requirement

We can ship it even without the bottom sheet. Many people who actually use Tron will already have TRX.

Follow up: https://github.com/daimo-eth/daimo/issues/1078

PR #1076, opened 2024-05-25
improve api deploy

dehli:
Updates cd so that it first checks whether an image exists with the git sha so that the build/push can be skipped. If it needs to deploy, it will build and deploy using docker, utilizing caching.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -86,7 +86,7 @@ jobs:
           fi
 
       - name: Deploy to ECS
-        run: aws ecs update-service --cluster $CLUSTER_NAME --service $SERVICE_NAME --force-new-deployment
+        run: aws ecs update-service --cluster $CLUSTER_NAME --service $SERVICE_NAME --force-new-deployment > /dev/null
 
   eas-deploy:
     if: github.ref == 'refs/heads/master'
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -84,9 +84,7 @@ jobs:
           fi
 
       - name: Deploy to ECS
-        run: |
-          pip install ecs-deploy
-          ecs deploy $CLUSTER_NAME $SERVICE_NAME --timeout 900
+        run: aws ecs update-service --cluster $CLUSTER_NAME --service $SERVICE_NAME --force-new-deployment
 
   eas-deploy:
     if: github.ref == 'refs/heads/master'
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -42,33 +42,43 @@ jobs:
         with:
           platforms: arm64
 
-      - name: Build, tag, and push docker image to Amazon ECR
+      - name: Check for existing image with GIT_SHA
         env:
           GIT_SHA: ${{ github.sha }}
-          REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           REPO: ${{ vars.AWS_ECR_API_REPO }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -37,24 +37,54 @@ jobs:
...
+            exit $status
+          else
+            echo "Image already tagged with \"$IMAGE_TAG.\""
+          fi
 
       - name: Deploy to ECS
-        run: |
-          pip install ecs-deploy
-          ecs deploy $CLUSTER_NAME $SERVICE_NAME --timeout 900
+        run: aws ecs update-service --cluster $CLUSTER_NAME --service $SERVICE_NAME --force-new-deployment
```
dcposch, line 87: 👍

.github/workflows/cd.yml
```diff
@@ -37,24 +37,54 @@ jobs:
...
-        uses: docker/setup-qemu-action@v2
+      - name: Check for existing image with GIT_SHA
+        env:
+          GIT_SHA: ${{ github.sha }}
+          REPO: ${{ vars.AWS_ECR_API_REPO }}
+        run: |
+          image_manifest=$(aws ecr batch-get-image --repository-name $REPO --image-ids imageTag=$GIT_SHA --query 'images[0].imageManifest')
+          echo "IMAGE_MANIFEST=$image_manifest" >> $GITHUB_ENV
+
+      - if: env.IMAGE_MANIFEST == 'null'
```
dcposch, line 48: why would IMAGE_MANIFEST be the string value "null" ?

.github/workflows/cd.yml
```diff
@@ -37,24 +37,54 @@ jobs:
...
+          set +e
+          output=$(aws ecr put-image --repository-name $REPO --image-tag $IMAGE_TAG --image-manifest "$IMAGE_MANIFEST" 2>&1)
+          status=$?
+          set -e
+
+          if [[ $status -ne 0 && ! "$output" =~ "ImageAlreadyExistsException" ]]; then
+            echo $output
+            exit $status
+          else
+            echo "Image already tagged with \"$IMAGE_TAG.\""
```
dcposch, line 83: this will print "Image already tagged..." even if the command was successful, right? when $status = 0

.github/workflows/cd.yml
```diff
@@ -37,24 +37,54 @@ jobs:
...
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - if: env.IMAGE_MANIFEST == 'null'
+        name: Build, tag, and push docker image to Amazon ECR
+        uses: docker/build-push-action@v5
         with:
-          platforms: arm64
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
dcposch, line 57: looks like `mode=max` is about caching intermediate (non-exported) Docker layers

do you know if our build has any of those?

dcposch (approved):
LGTM after comments

.github/workflows/cd.yml
dehli: Correct. Since we're storing the result in an environment variable it gets cast to a string which is why we need to check for the `'null'` string value.

.github/workflows/cd.yml
dehli: good catch! meant to have that nested under the non-zero catch. will address

.github/workflows/cd.yml
dehli: I don't think so. Went with max just because it could cache more but probably doesn't make a difference for the api image.

Issue #964, opened 2024-04-04
Android generate device key bug "null"


dcposch:
## Summary

Version: 1.8.0
Platform: Android 12 / Pixel 3

Following observed occasionally, eg after logging out of an account and trying to create a testnet account.

Click Generate.

## Expected behavior

Asks for lock pattern, generates key, proceeds

## Actual behavior

In this case, the system lock pattern didn't show like it should have, it just went directly to this:

<img width=300 src="https://github.com/daimo-eth/daimo/assets/169280/510b4c30-3863-487d-902c-ac361d532080" />

## Steps to reproduce

Unknown, no reliable repo.

dcposch:
fixed in #1081

PR #1081, opened 2024-05-28
android: fix null error

nalinbhardwaj:
bad onboarding bug on android, no key is generated in the use existing flow at the moment:

> 2024-05-27T21:50:18.630Z ERR ExpoEnclaveSign: Encountered an exception while calling native method: Exception occurred while executing exported method sign on module ExpoEnclave: null


dcposch (approved):
LGreatTM

reproduced bug on Pixel 3 (log in > log out > try "use existing" again)

confirmed fix on same phone

dcposch:
**fixes #964**

Issue #1082, opened 2024-05-28
Eth transfer no swap notif


andrewliu08:
Tried sending eth from my metamask to daimo account. Didn't receive a notif to convert the eth to usdc via uniswap

Tx hash
https://basescan.org/tx/0x6ff507b2f1fc8e28210c9c1eeee98d7f61be0ccfe8251c5441b1d769e52c01e2

My daimo address:
0x6af35df65594398726140cf1bf0339e94c7a817f

dcposch:
Looks like this bug is back.
- ETH transfer to https://daimo.com/l/account/leftcivet44
- Transfer success: https://basescan.org/tx/0x2c43b0eec214e4e390f51eee6667636167467e29ba2f6d6868fedbbbc3f54b29
- Transfer appears correctly in `index.daimo_transfers`
- No pending swaps for leftcivet44

PR #1035, opened 2024-04-27
shovel: new eth transfers feature

nalinbhardwaj:
supersede https://github.com/daimo-eth/daimo/pull/1029.

Upgrades shovel to include ETH transfers with multiple RPC urls (fallback for when non-ratelimited RPC fails).

PR #1089, opened 2024-05-28
api: disable ethIndexer for now, bugs

nalinbhardwaj:
Looks like some weird bugs due Quicknode breaking and not working well with the `updateLastProcessedCheckStale` which doesn't account for it -- will just disable for now and bring it back in #1035 properly, additionally fixing #1082.

Tested API still works fine.


dcposch (approved):
LGTM

Issue #1088, opened 2024-05-28
Quicker finalization of transaction in UI


nounderline:
It might take up to 5 seconds for the app to display a finalized transaction. We want to bring it down to ~2s (2s Base block + network jitter + indexing delay.)

Consider following process diagram of app transaction synchronization:

![Daimo sync drawio (1)](https://github.com/daimo-eth/daimo/assets/765841/42559bfa-656f-403f-a35f-0629d4243ea9)

- We do not consider network conditions, device load, and platform scheduling here.
- We assume transaction is included in next block.
- We assume new block is instantly visible across Base network and there is no indexing delay.

Even with idealized conditions, user may still wait for up to 4s for transaction to appear finalized in the UI. Main reason for this delay is [resync()](https://github.com/daimo-eth/daimo/blob/15044c9b5a4945479d9f021619d09ca0b46ee3f3/apps/daimo-mobile/src/sync/sync.ts#L102) throttling and polling updates (which happens every 1s.)

We can minimize this time by pushing information about new block to an app and requesting a sync immediately when a new block is indexed.

We can further minimize the delay by sending all necessary information about the transaction in an event payload and avoid sync request saving additional network roundtrip. This may save 100-300ms for a mobile user that is physicailly distant from API server (think US-Turkey.)

We could add a server-sent event endpoint that emits a new event when we index it. As we index onchain data to Postgres via Shovel we can either [Shovel integration](https://www.indexsupply.com/shovel/docs/#config-integrations-notification) or use Postgres triggers directly to emit notification that will be consumed via this endpoint.

Another alternative could be push notification but they are more platform-dependent and, in our case, relay on three providers we do not control: Expo and Google/Apple. Since mobile push notifications may be delayed or dropped altogether, I'd avoid push notifications for any in-app information control.

dcposch:
Details from Notion:

### Part 1 - Fast Transactions

One of the most important metrics in Daimo is **transaction speed**. I send a transfer: the transfer shows as pending quickly, then as confirmed in a rollup block several seconds later.

Today, this is too slow. I’ve experienced 5-10 seconds. For this project, we’ll measure this latency and then reduce it as far as possible. The theoretical minimum is 1s average, 2s worst-case driven by Base’s 2-second block time.

We can split userop latency into three parts. Measure all three (before), then reduce as far as possible.

[...]

Issue #1098, opened 2024-05-31
Speed up indexing transfers


nounder:
It takes 1-3s for transfers to be indexed by Shovel after they have been produced on Base.

![image](https://github.com/daimo-eth/daimo/assets/765841/637bfe52-4235-49a7-b9dd-67c4cc6dafea)

A histogram of duration it takes for a block to be indexed in Shovel staging database after block has been produced on Base Sepolia as advertised by [Basescan](https://sepolia.basescan.org/block/10701860). Indexing time is based on [shovel.task_updates](https://www.indexsupply.com/shovel/docs/#schema) for 'transfers' indexer, ie. the time transfer records are committed to the database. For full methodology and code see [this notebook](https://gist.github.com/nounderline/6dd490ea541b680087e70aa719556ef3).

We can reduce the delay by bringing Base node, database, Shovel, and RPC server closer together. Since we're using timestamp from Basescan as a baseline it would be good to find a definition for it.

<details><summary>Python Notebook</summary>
<p>

```python
# %%
!pip install httpx polars connectorx pyarrow plotly pandas kaleido

# %%
from collections import namedtuple
import os
import httpx
import polars as pl
from functools import lru_cache
from datetime import datetime

BlockInfo = namedtuple(
    "BlockInfo",
    "blockNumber timeStamp blockMiner blockReward uncles uncleInclusionReward",
)


@lru_cache
def fetch_block(block_number, api_key="YourApiKeyToken"):
    url = f"https://api-sepolia.basescan.org/api?module=block&action=getblockreward&blockno={block_number}&apikey={api_key}"
    response = httpx.get(url)
    data = response.json()
    result = data.get("result", {})
    timestamp = datetime.utcfromtimestamp(int(result.get("timeStamp")))
    return BlockInfo(
        blockNumber=int(result.get("blockNumber")),
        timeStamp=timestamp,
        blockMiner=result.get("blockMiner"),
        blockReward=result.get("blockReward"),
        uncles=result.get("uncles"),
        uncleInclusionReward=result.get("uncleInclusionReward"),
    )


def fetch_block_sample(last_block, length):
    return [fetch_block(n) for n in range(last_block - length, last_block)]


def read_recent_shovel_indexed_transfer(limit=100):
    return pl.read_database_uri(
        f"""
select distinct src_num, insert_at
from shovel.task_updates
where
ig_name = 'transfers'
order by src_num desc
limit {limit}
""",
[...]

PR #1102, opened 2024-06-03
Realtime push for account changes

### LINKED ISSUE #1144: Store API indexed state in database
API server relies on Watcher to query and index data such as block height, ENS registry and Daimo names. This has some problems:

1. Server startup slow as it needs to index all this data before serving requests.
2. Scaling horizontally is quite a pickle because each instance would have it's own state and return different results.

Having a single instance consumer that indexes the state and pushes it to a database shared across all API instances could be a solution here.

nounderline:
WIP. Closes https://github.com/daimo-eth/daimo/issues/1088 closes #1144 

- Replace polling for account history with subscription via WebSockets.
- Add `onAccountUpdate` TRPC subscription that pushes new `AccountHistoryResult` on new transactions;
- or sends a new `AccountHistoryResult` every 10s to update exchange rates and related values.
- Add `DBNotification` interface to listen to Postgres notifications that inform about newly added transactions by Shovel.
- Replace polling for API `Watcher` with `DBNotification` to enable internal state change new new block is indexed.


nounderline:
Since I have no access to staging Shovel instance I created a trigger myself:

```
-- Trigger function that executes after an insert
CREATE OR REPLACE FUNCTION after_insert_tx_notify()
RETURNS TRIGGER AS $$
DECLARE
  latest_num INTEGER;
BEGIN
  SELECT num INTO latest_num FROM shovel.latest WHERE src_name = 'baseSepolia';
  PERFORM pg_notify('newBlock', json_build_object('number', latest_num)::text);
  RETURN NULL;
END;
$$ LANGUAGE plpgsql;

-- Trigger on transfers table
CREATE TRIGGER after_insert_tx_notify
AFTER INSERT ON transfers
FOR EACH STATEMENT
EXECUTE FUNCTION after_insert_notify();
```

Ultimately we want to use [Shovel notifications](https://www.indexsupply.com/shovel/docs/#notifications) if I we managed to improve indexing performance in https://github.com/daimo-eth/daimo/issues/1098. If not, we may connect directly to the RPC. Will need more information about current infrastructure to decide on best course here (I messaged @dcposch for details on Telegram.)

apps/daimo-mobile/src/logic/trpc.tsx
```diff
@@ -55,69 +63,90 @@ function chooseChain<T>({
...
+  const method = init.method || "GET";
+  console.log(`[TRPC] ${method} ${func} ${ret.status} in ${ms}ms`);
+
+  return ret;
+};
+
+function getTRPCOpts(
+  daimoChain: DaimoChain
+): CreateTRPCClientOptions<AppRouter> {
+  const url = `http://localhost:2022`;
```
dcposch, line 119: what's localhost:2022 ?

apps/daimo-mobile/src/logic/trpc.tsx
```diff
@@ -55,69 +63,90 @@ function chooseChain<T>({
...
-          return ret;
+      // call subscriptions through websockets and the rest over http
+      splitLink({
+        condition(op) {
+          return op.type === "subscription";
         },
+
+        true: wsLink({
+          client: createWSClient({
+            url,
```
dcposch, line 131: nice!

how does the 1 websocket URL (vs 2 for mainnet/testnet) work?

i have an idea for how we can simplify in a followup PR --requires some API surgery

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -42,7 +36,45 @@ export function startSync() {
...
+
+    const sub = rpcFunc.onAccountUpdate.subscribe(
+      {
+        address: account.address,
+        sinceBlockNum: account.lastFinalizedBlock,
+      },
+      {
+        onData: (data) => {
+          console.log("received account update", data);
+          applySync(account, data, false);
```
dcposch, line 67: 🔥

packages/daimo-api/src/db/notifications.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  async close() {
+    if (this.client) {
+      await this.client.end();
+
+      this.channels.clear();
+    }
+  }
+
+  async #setupListener() {
```
dcposch, line 48: how does `#method` work vs `private method` ?

dcposch (commented):
Looks great so far!

apps/daimo-mobile/src/logic/trpc.tsx
nounderline: Ah, that's a local server with some modifications to make testing easier (no ready guard and no indexing)

packages/daimo-api/src/db/notifications.ts
nounderline: private method is typescript the compiler thing, #method is javascript the language thing. the latter came later that's why we have both of them in ts now. also, #method make sure method is private in the runtime. 

can move to `private method` for consistency across code base.

apps/daimo-mobile/src/logic/trpc.tsx
nounderline: I haven't considered that. Is it the case that we use both mainnet and testnet in the app?

packages/daimo-api/src/db/notifications.ts
nounderline: Done in a1a561f970c8c5d7ac0f0cad889d8de815785ce7

nounderline:
Pushing account history works great now!

Remaining work:

 - Ensure websockets connection retries on network/server failure.
 - Think about pull-to-refresh and how it plays with push-based sync. Remove it altogether maybe?
 - Limit getAccountHistory data emission. Now that we have stateful connection we can push time-sensitive data (ie. transactions) while omitting data like exchange rates.

apps/daimo-mobile/src/logic/trpc.tsx
```diff
@@ -56,69 +66,118 @@ function chooseChain<T>({
...
+  const { pathname } = url;
+  const func = pathname.split("/").slice(-1)[0] as keyof AppRouter;
+  const timeout = (() => {
+    if (func === "deployWallet") return 60_000; // 1 minute
+    else return 10_000; // default: 10 seconds
+  })();
+  console.log(`[TRPC] fetching ${url}, timeout ${timeout}ms`, init);
+  const controller = new AbortController();
+  const timeoutID = setTimeout(() => {
+    console.log(`[TRPC] timeout after ${timeout}ms: ${input}`);
```
dcposch, line 97: i think this will always print `[TRPC] timeout ...`, even if the function didn't time out

apps/daimo-mobile/src/logic/trpc.tsx
```diff
@@ -56,69 +66,118 @@ function chooseChain<T>({
...
+  });
+
+  let daimoLink = httpBatchLink({
+    url: chooseChain({
+      daimoChain,
+      mainnet: apiUrlMainnetWithChain,
+      testnet: apiUrlTestnetWithChain,
+    }),
+    fetch: customTRPCfetch,
+  });
```
dcposch, line 135: you can deduplicate a bit here: 

```ts
const daimoHttpLink = httpBatchLink({ url, fetch: customTRPCFetch })
```

then below

```
if (typeof WebSocket == null) {
  return daimoHttpLink;
}

// splitLink goes here
```

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -7,127 +7,124 @@ import {
...
-//   ...occasionally otherwise
-export function startSync() {
-  console.log("[SYNC] APP LOAD, starting sync");
-  maybeSync(true)
-    .then((status) => {
-      if (status === "failed") {
-        updateNetworkState(() => {
-          return { status: "offline", syncAttemptsFailed: 1 };
-        });
+class SyncManager {
```
dcposch, line 20: 👍

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -7,127 +7,124 @@ import {
...
+  manager = getAccountManager();
+  currentAccount: Account | null = null;
+  // for tracking retry timeout
+  retryTimeout: any = undefined;
+
+  _trpcUnsubscribe: (() => void) | null = null;
```
dcposch, line 26: can use `private ` instead of _

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -7,127 +7,124 @@ import {
...
+        address: account.address,
+        sinceBlockNum: 0,
+      },
+      {
+        onStarted: () => {
+          updateNetworkStateOnline();
+        },
+
+        onData: (data) => {
+          this.manager.transform((a) => applySync(a, data, false));
```
dcposch, line 55: 🔥

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -7,127 +7,124 @@ import {
...
+    this.currentAccount = account;
+
+    const sub = rpcFunc.onAccountUpdate.subscribe(
+      {
+        address: account.address,
+        sinceBlockNum: 0,
+      },
+      {
+        onStarted: () => {
+          updateNetworkStateOnline();
```
dcposch, line 51: will this automatically reconnect if network goes down and comes back?

apps/daimo-mobile/src/sync/sync.ts
```diff
@@ -7,127 +7,124 @@ import {
...
 
-  // Sync faster for 1. pending ops or expired swaps, and 2. recently-failed sync
-  if (hasPendingOps(account) || hasCacheExpiredSwaps(account)) {
-    intervalS = 1;
-  }
+  // create small delay to let interface render becuase this callback
+  // run right after getting data so the interface still got to adapt
+  setTimeout(() => {
+    SplashScreen.hideAsync();
+  }, 300);
```
dcposch, line 103: dont know if we need this here. all good, fixed in followup PR

packages/daimo-api/src/db/notifications.ts
```diff
@@ -0,0 +1,91 @@
...
+  chainConfig.daimoChain + "-daimo-transfers";
+
+const IGNORED_EVENTS = new Set(["newListener", "removeListener"]);
+
+/**
+ * Connects to Postgres and emits LISTEN/NOTIFY events.
+ */
+export class DBNotifications extends EventEmitter {
+  client: Client;
+  channels: Set<string>;
```
dcposch, line 16: add 1line comment explaining what channels are

packages/daimo-api/src/db/notifications.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!this.isConnected) {
+      await this.#setupListener();
+    }
+
+    if (!this.channels.has(channel)) {
+      this.channels.add(channel);
+
+      try {
+        await this.client.query(`LISTEN "${channel}"`);
+      } catch (e) {
```
dcposch, line 77: add console.log right before LISTEN / UNLISTEN

packages/daimo-api/src/server/router.ts
```diff
@@ -113,13 +118,19 @@ export function createRouter(
 
   // CORS for the web app
   const corsMiddleware = trpcT.middleware(async (opts) => {
-    opts.ctx.res.setHeader("Access-Control-Allow-Origin", "*");
+    // cannot set headers when connecting via websockets
```
dcposch, line 121: 👍

packages/daimo-api/src/server/router.ts
```diff
@@ -113,13 +118,19 @@ export function createRouter(
...
+    }
+
     return opts.next();
   });
 
   // Don't serve requests until we're ready.
   // This avoids confusing UI state in local development.
   const readyMiddleware = trpcT.middleware(async (opts) => {
+    // Don't serve requests until we're ready.
+    // This avoids confusing UI state in local development.
```
dcposch, line 133: remove, duplicated

packages/daimo-api/src/server/router.ts
```diff
@@ -624,6 +635,90 @@ export function createRouter(
...
+          address: zAddress,
+          inviteCode: z.string().optional(),
+          sinceBlockNum: z.number(),
+        })
+      )
+      .subscription(async (opts) => {
+        const { address, inviteCode } = opts.input;
+        // how often to send updates regardless of new transfers
+        // useful to update exchange rates and others.
+        const refreshInterval = 10_000;
```
dcposch, line 650: 👍

packages/daimo-api/src/server/router.ts
```diff
@@ -624,6 +635,90 @@ export function createRouter(
...
+          };
+
+          const intervalTimer = setInterval(() => {
+            // interval concided with new block. let's skip this one.
+            if (getAccountHistoryPromise) {
+              return;
+            }
+
+            pushHistory(false);
+          }, refreshInterval);
```
dcposch, line 710: this is quite tricky... if I'm understanding correctly:

- interval fetch skipped if we're currently doing an event-based fetch
- however, event-based fetch is *not* skipped if we're doing an interval fetch

in that second case, both can run concurrently.

packages/daimo-api/src/server/router.ts
```diff
@@ -624,6 +635,90 @@ export function createRouter(
...
+              nameReg,
+              keyReg,
+              paymaster,
+              db
+            );
+
+            getAccountHistoryPromise
+              .then((history) => {
+                // we can have concurrent requests. discard those that arrived too late
+                if (history.lastBlock < lastEmittedBlock) {
```
dcposch, line 682: `<=`

packages/daimo-api/src/server/router.ts
```diff
@@ -624,6 +635,90 @@ export function createRouter(
...
+                  return;
+                }
+
+                if (onlyOnNewTransfers && history.transferLogs.length === 0) {
```
dcposch, line 686: what if, say, a keyAdded event arrives > we resync. no new transfers, but we have a new key.

won't this line discard the sync result?

dcposch (commented):
LGreatTM in general. left comments

apps/daimo-mobile/src/logic/trpc.tsx
dcposch: fixed

apps/daimo-mobile/src/logic/trpc.tsx
dcposch: done

packages/daimo-api/src/db/notifications.ts
dcposch: done

packages/daimo-api/src/server/router.ts
dcposch: done

dcposch:
Looking forward to ship. 

- Watcher no longer has an interval tick, only a subscription to DB_EVENT_DAIMO_TRANSFERS. For other events, like adding a signing key, how does the watcher learn about them?
- "Ultimately we want to use [Shovel notifications](https://www.indexsupply.com/shovel/docs/#notifications) if I we manage to improve indexing performance in https://github.com/daimo-eth/daimo/issues/1098" Why does this depend on #1098 ? The way we're doing it currently is still downstream of shovel > making shovel faster is a good goal but I think using Shovel Notifications seems correct either way. LMK if there's a reason *not* to use Shovel notifications instead of triggers.

packages/daimo-api/src/server/router.ts
nounderline: Yes, your interpretation is correct. In both cases we favor event-triggered sync over interval one because it is guaranteed to have latest data.
For 1) we skip interval because data was requested before Shovel committed data to Postgres. For 2) we do don't do interval-triggered query because there is already event-triggered query running, which will have the latest data.

packages/daimo-api/src/server/router.ts
nounderline: That's why there is interval sync on top of event-triggered sync that updates keys, exchange rates, linked accounts, named accounts, suggested actions, invites, notifications requests, and proposed swaps.

This way transfers are pushed in real time transfers while maintaining existing pull behavior for updating full state without major refactor.

packages/daimo-api/src/server/router.ts
nounderline: Aha! I just noticed an unexpected edge case for case 2: when interval sync coincides with event-triggered sync and there is no new transfer for that block, user will have to wait for another interval sync to get updated about other `AccountHistoryResult` changes (namely keys, exchange rates, linked accounts, named accounts, suggested actions, invites, notifications requests, and proposed swaps.)

Thanks for @dcposch!

nounderline:
@dcposch 

> Watcher no longer has an interval tick, only a subscription to DB_EVENT_DAIMO_TRANSFERS. For other events, like adding a signing key, how does the watcher learn about them?

`DB_EVENT_DAIMO_TRANSFERS` is emitted on new block. Since [Shovel updates task tables atomically](https://www.indexsupply.com/shovel/docs/#notifications), updates to remaining tables will committed to database before Postgres emits that event.


> LMK if there's a reason not to use Shovel notifications instead of triggers.

Shovel send individual notifications for each row in a form of a row tuple (example: `$block_num,$amount,$address`). 

Custom trigger we use here sends a notification only when Shovel inserts and successfully indexes a block. This provides us with one notification per block instead of many notifications for every row added across all tracked tables.

Shovel notifications are sent when [new row is inserted](https://www.indexsupply.com/shovel/docs/#config-integrations-notification). If there are no rows added, there is no way to know about new blocks using this feature.

This simplifies implementation for API `Watcher` and `onAccountUpdate` TRPC subscription because they are only notified  on new block once without requiring them to filter through multiple notifications sent for each row.

packages/daimo-api/src/shovel/watcher.ts
```diff
@@ -109,7 +113,7 @@ export class Watcher {
   // Watches shovel for new blocks, and indexes them.
   // Skip indexing if it's already indexing.
   async watch() {
-    setInterval(async () => {
+    this.notifications.on(DB_EVENT_DAIMO_NEW_BLOCK, async () => {
```
dcposch, line 116: 🔥

dcposch (approved):
LGTM!

Issue #858, opened 2024-03-15
move testnet links to their own domain


dcposch:
## Goal

Clean separation, no ambiguity between mainnet and testnet accounts.

## Proposed fix.

`https://daimo.com/invite/abc123` is a mainnet deeplink

`https://test.daimo.xyz/invite/testnet` is a test deeplink

Pairs well with #620

PR #1106, opened 2024-06-04
web: getEnvVars with typecheck

dcposch:
## Strongly typed environment variables

This PR cleans up `daimo-web` and `daimo-api`, along with their shared dependency `daimo-common`. The goal is the following for all executables:
* List all environment variables in a central place
* On startup, validate that all required vars are present and all present vars validate
* Log environment for easier debugging

I had to do a few cleanups in order to ship this.
- `daimo-common` no longer relies on environment variables at all. Only executables have env vars--libraries should have any configuration they need passed in.
- Deleted `DAIMO_DOMAIN`. Its meaning wasn't well defined, and we set it to `daimo.com` in both staging and production. Its original purpose is superseded, will be fixed in #858 
- Deleted unused env vars
- Added documentation in `.env.example`

The goal is for every executable to have an `env.ts` that defines inputs, along with `.env.example` to make it as fast as possible to start development.

cc @andrewliu08 @nounderline


packages/daimo-api/src/env.ts
```diff
@@ -1,10 +1,64 @@
...
+    .string()
+    .optional()
+    .transform((s) => s === "true"),
+  // API operator internal notifications: Clippy, Slack bot
+  CLIPPY_WEBHOOK_URL: z.string().optional().default(""),
+  // Monitoring: Honeycomb
+  HONEYCOMB_API_KEY: z.string().optional().default(""),
+  // Monitoring: Sentry
+  SENTRY_DSN: z.string().optional().default(""),
+};
```
dcposch, line 52: going forward all programs should have one of these + no use of `process.env` outside of `env.ts`

PR #1108, opened 2024-06-04
dev: fix build, fix exchange rates

dcposch:
## Fix build

We're in dependency hell. Upgrading packages even slightly--latest Typescript, minor version bump in React, etc--causes a cascade of bundler errors. Some affect the web build (Next), others the mobile build (expo/metro).


dcposch:
## Fix exchange rates
Follow up to #1096 
Deploying to fix the exchange rate calculator

dcposch:
## Disabled tests

cc @nalinbhardwaj since I think you ran into a similar issue before

shipping this PR now to fix the build, but I'd like to re-enable these two `AccountSendUserop` tests if we can make them non-flaky.

https://github.com/daimo-eth/daimo/pull/1108/commits/62664846c571c25cff6d0fd2f2c8ada9fe43dec6

PR #1119, opened 2024-06-06
add bot & clippy to cd

dehli:
Couldn't find a nice way to share the common steps between jobs so just repeated them. Also adds `ts-node` as a devDependency to clippy matching the bot setup (`tsx` was failing to run on the docker image).

### DIFF
--- a/apps/daimo-clippy/Dockerfile
+++ b/apps/daimo-clippy/Dockerfile
@@ -1,0 +1,20 @@
+# Use the official Node.js 20 image.
+FROM node:20
+
+# Set the working directory inside the container
+WORKDIR /usr/src/app
+
+# Copy app code
+COPY . .
+
+# Install dependencies
+RUN npm ci
+
+# Build app
+RUN npm run build:clippy
+
+# Expose the port the app runs on
+EXPOSE 4200
+
+# Start app
+CMD ["npm", "start", "--prefix", "./apps/daimo-clippy"]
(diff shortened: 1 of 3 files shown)


apps/daimo-clippy/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN npm ci
+
+# Build app
+RUN npm run build:clippy
+
+# Expose the port the app runs on
+EXPOSE 4200
+
+# Start app
+CMD ["npm", "run", "start:v2", "--prefix", "./apps/daimo-clippy"]
```
dcposch, line 20: why do we need a start:v2 ?

apps/daimo-clippy/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "1.0.0",
   "main": "src/index.ts",
   "scripts": {
-    "dev": "echo Use 'npm start' for Clippy dev",
+    "dev": "ts-node-dev src/index.ts",
     "start": "tsx src/index.ts",
+    "start:v2": "ts-node src/index.ts",
```
dcposch, line 8: ah, i see. what was the tsx error?

dcposch (approved):
LGreatTM after comments

overall tsx is better than ts-node, so if possible i'd prefer to run that way.
what was the error?

apps/daimo-clippy/package.json
dehli: Weird, just rebuilt and it's no longer giving an error. Wonder if it might have been b/c I had copied node_modules over at first. Undid these changes though.

dehli: Weird, just rebuilt and it's no longer giving an error. It probably is b/c I had copied node_modules over at first. Undid the changes 👍

PR #1127, opened 2024-06-08
api: ethIndexer caching changes

kayleegeorge:
Only change cache when there is a balance difference instead of updating cache for all accounts every index.


dcposch (approved):
LGTM

PR #1129, opened 2024-06-09
mobile, api: add request memos

dcposch:
## Sender sends request

Screenshot 2024-06-09 at 13 05 42
Screenshot 2024-06-09 at 13 10 56

## Recipient fulfills

image
image
image
image
image

## Sender sees fulfilled transfer

image
image

### DIFF
--- a/packages/daimo-api/src/api/createRequestSponsored.ts
+++ b/packages/daimo-api/src/api/createRequestSponsored.ts
@@ -4,20 +4,23 @@ import { Address, Hex, stringToHex } from "viem";
 
 import { RequestIndexer } from "../contract/requestIndexer";
 import { ViemClient } from "../network/viemClient";
+import { PaymentMemoTracker } from "../offchain/paymentMemoTracker";
 
 interface RequestV2Input {
   idString: string;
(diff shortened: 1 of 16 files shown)


packages/daimo-api/src/api/createRequestSponsored.ts
```diff
@@ -40,5 +46,10 @@ export async function createRequestSponsored(
     args: [id, recipient, BigInt(amount), metadata],
   });
 
+  // Store memo, if applicable
+  if (memo != null) {
+    paymentMemoTracker.addMemo(requestTxHash, memo);
```
dcposch, line 51: one slight ugliness: request memos are tracked by *tx hash*, not op hash

i think this is OK:
* txhash and ophash are non-overlapping
* for permissionless request creation, it'd be an op anyway
* it's currently a tx because we want to allow people to send requests without onchain auth = without FaceID

(separately, session key auth in the API will make that setup less resistant to abuse. coming soon. cc @kayleegeorge )

dcposch:
shipping, cc @kayleegeorge @nalinbhardwaj for post review

PR #1137, opened 2024-06-11
api: fix Viem transactions

dcposch:
add VSCode launch configuration to debug daimo-api

workaround Viem error-printing bug


kayleegeorge (approved):
when logging breaks the API :( LGTM

PR #1142, opened 2024-06-12
api: faucet protection

dcposch:


packages/daimo-api/src/env.ts
```diff
@@ -21,7 +21,6 @@ const zEnv = {
   EXCHANGE_RATES_URL: z.string().startsWith("https://"),
   // Anti-spam faucet API URL
   DAIMO_FAUCET_API_URL: z.string().optional().default(""),
-  DAIMO_FAUCET_API_KEY: z.string().optional().default(""),
```
nalinbhardwaj, line 24: intentional? also drop from https://github.com/daimo-eth/daimo-infra/commits/master/ if so

packages/daimo-api/src/server/antiSpam.ts
```diff
@@ -0,0 +1,46 @@
...
+  private static url = getEnvApi().DAIMO_FAUCET_API_URL;
+
+  // Check whether a request is eligible for invite link faucet.
+  public static async shouldSendFaucet(reqInfo: any): Promise<boolean> {
+    // Staging = no faucet API = always allow
+    if (AntiSpam.url === "") return true;
+
+    // Prod = query API to avoid sending invite faucet to spammers / farmers
+    const faucetUrl = new URL("/faucet", AntiSpam.url);
+    const resp = await AntiSpam.tryQuery<any>(faucetUrl, reqInfo);
```
nalinbhardwaj, line 14: nit, would type

packages/daimo-api/src/server/antiSpam.ts
```diff
@@ -0,0 +1,46 @@
...
+    return sendFaucet;
+  }
+
+  // Check whether a request can use the Daimo API at all.
+  public static async shouldServeAPI(reqInfo: any): Promise<boolean> {
+    // Staging = no faucet API = always allow
+    if (AntiSpam.url === "") return true;
+
+    const faucetUrl = new URL("/api-protection", AntiSpam.url);
+    const resp = await AntiSpam.tryQuery<any>(faucetUrl, reqInfo);
```
nalinbhardwaj, line 28: same, would type response manually

nalinbhardwaj (approved):
lgtmbc

packages/daimo-api/src/env.ts
dcposch: wasn't actually enforced before. can add.

Issue #1109, opened 2024-06-04
Use websockets in Shovel


nounderline:
Current Shovel configuration polls for a new block by sending individual RPC `eth_getBlockByNumber` calls every 1 second. This increases indexing latency as described in #1098.

https://github.com/daimo-eth/daimo/blob/e94ed904a4204ace5fc4d99c0133cbd87f389ac0/packages/daimo-contract/src/shovel/config.json#L4-L12

We should use WebSockets RPC endpoint instead to enable low-latency indexing. To do so, `ws_url` is to be added to the source [configuration](https://www.indexsupply.com/shovel/docs/#config-eth-sources-ws-url).

> eth_sources[].ws_url
> An optional URL that points to a Websocket JSON RPC API. If this URL is set Shovel will use the websocket to get the latest block instead of calling eth_getBlockByNumber. If the websocket fails for any reason Shovel will fallback on the HTTP based API.

PR #1150, opened 2024-06-13
Use Websocket in Shovel for realtime block ingestion

nounderline:
Simple but effective change.

```
eth_sources[].ws_url
An optional URL that points to a Websocket JSON RPC API. If this URL is set Shovel will use the websocket to get the latest block instead of calling eth_getBlockByNumber. If the websocket fails for any reason Shovel will fallback on the HTTP based API
```

More: https://www.indexsupply.com/shovel/docs/#config-eth-sources-ws-url

This requires adding CHAIN_RPC_WS_URL env var to Shovel service.

closes https://github.com/daimo-eth/daimo/issues/1109

### DIFF
--- a/packages/daimo-contract/src/shovel/config.json
+++ b/packages/daimo-contract/src/shovel/config.json
@@ -6,6 +6,7 @@
       "name": "$CHAIN_NAME",
       "chain_id": "$CHAIN_ID",
       "url": "$CHAIN_RPC_URL",
+      "ws_url": "$CHAIN_RPC_WS_URL",
       "batch_size": 100,
       "concurrency": 4
     }
@@ -1210,3 +1211,4 @@
     }
   ]
 }
+


dcposch (approved):

packages/daimo-contract/src/shovel/config.json
```diff
@@ -6,6 +6,7 @@
       "name": "$CHAIN_NAME",
       "chain_id": "$CHAIN_ID",
       "url": "$CHAIN_RPC_URL",
+      "ws_url": "$CHAIN_RPC_WS_URL",
```
dcposch, line 9: proposal

```
"ws_url": "$CHAIN_RPC_URL_WS",
"urls": ["$CHAIN_RPC_URL", "$CHAIN_RPC_URL_BACKUP"],
```

^ i'll make the corresponding change in infra

dcposch: actually that second part is done in https://github.com/daimo-eth/daimo/pull/1035

shipping

PR #1152, opened 2024-06-13
Minimize splash screen delay

### LINKED ISSUE #1111: Remove splash screen delay
Splash screen is manually hidden 300ms after sync either returns or throws:

https://github.com/daimo-eth/daimo/blob/e94ed904a4204ace5fc4d99c0133cbd87f389ac0/apps/daimo-mobile/src/sync/sync.ts#L42-L44

This increases time-to-interaction time by at least 1/3 of a second. We can make the app feel start up faster by showing UI immediately with [cached account data](https://github.com/daimo-eth/daimo/blob/e94ed904a4204ace5fc4d99c0133cbd87f389ac0/apps/daimo-mobile/src/logic/accountManager.ts#L99-L100) with visible sync indicator.

nounderline:
Moves SplashScreen management to Navigator to avoid initial blank state. Changed delay from 300ms to 32ms (duration of approx 2 ticks for UI work.)

closes #1111
to be merged after #1102


apps/daimo-mobile/src/view/TabNav.tsx
```diff
@@ -153,6 +156,13 @@ function MainTabNavigator() {
 export function TabNav() {
   const { account, keyInfo } = useAccountAndKeyInfo();
 
+  // Remove splash screen
+  useEffect(() => {
+    setTimeout(() => {
+      SplashScreen.hideAsync();
+    }, 32); // just wait for ~2 ticks for states to update
```
dcposch, line 163: wary of timeouts, i think we could this more directly eg 

```ts
useEffect(() => {
  const ageS = now() - account.lastBlockTimestamp
  // if ageS < 60 = 1min, and splash screen is visible, hide it
}, [account.lastBlockTimestamp]);
```

apps/daimo-mobile/src/App.tsx
```diff
@@ -33,19 +29,6 @@ export default function App() {
...
-
-  // Remove splash screen
-  useEffect(() => {
-    const nowS = now();
-    if (account == null || nowS - account.lastBlockTimestamp < 60 * 10) {
-      console.log(`[APP] removing splash now, first render`);
-      SplashScreen.hideAsync();
-    } else {
-      console.log(`[APP] removing splash after sync or 2s, whichever is first`);
-      setTimeout(() => SplashScreen.hideAsync(), 2000);
```
dcposch, line 45: hmmm now that i'm seeing this, what if we just delete the redundant 300ms timeout below and reduce the max-wait-time here?

if this `useEffect` is working correctly, it should already hide the splash screen as soon as the account is updated

dcposch (commented):
LGTM after comments

apps/daimo-mobile/src/App.tsx
nounderline: This will result in blank screen because the splash screen is hidden before navigator is rendered. That's why `SplashScreen.hideAsync()` is in navigator now

dcposch: the purpose of the using `SplashScreen.preventAutoHideAsync()` was to wait for the initial `getAccountHistory()` to return, so that we don't flash an incorrect/stale balance

dcposch (approved):

PR #1124, opened 2024-06-07
mobile: send any stablecoin UI

kayleegeorge:
Choose which stablecoin you want to send (USDC, DAI, USDT) + ETH UI. Uses on chain swap quoter to pull swap routes/quotes. Previews how much of send coin the recipient will receive

Screenshot 2024-06-21 at 1 28 56 PM

Screenshot 2024-06-24 at 11 40 47 AM

Cannot choose a different send coin when sending money to another Daimo account:

Screenshot 2024-06-21 at 1 35 02 PM

PR #1162, opened 2024-06-17
api: switch out fetch route to onchain oracle

kayleegeorge:
Fetches a swap route from the on-chain oracle rather than using uniswap.


packages/daimo-common/src/foreignCoin.ts
```diff
@@ -1,4 +1,5 @@
-import { Address, formatUnits } from "viem";
+import { WETH9 } from "@uniswap/sdk-core";
```
nalinbhardwaj, line 1: I think you could drop all the changes in this file for now and keep them for #1124

packages/daimo-api/src/api/getSwapRoute.ts
```diff
@@ -0,0 +1,139 @@
...
+  fromAccount: EAccount;
+  toAddr: Address;
+  vc: ViemClient;
+  chainId: number;
+}) {
+  const amountIn: bigint = BigInt(amountInStr);
+  const swapQuote = await vc.publicClient.readContract({
+    address: "0xD49Ad4D3945A655C508498571cFf092D261bf860", // TODO: config
+    abi: parseAbi([
+      "function quote(uint128 amountIn, address tokenIn, address tokenOut) public view returns (uint256 amountOut, bytes memory swapPath)",
```
nalinbhardwaj, line 42: Should include contract Solidity and use wagmi ABI generation as we talked about

packages/daimo-api/src/api/getSwapRoute.ts
```diff
@@ -0,0 +1,139 @@
...
+      amountIn,
+      amountOutMinimum,
+    };
+    callData = SwapRouter02.INTERFACE.encodeFunctionData("exactInput", [
+      exactInputParams,
+    ]);
+  }
+
+  if (!callData) return null;
+  const execCallData = SwapRouter02.INTERFACE.encodeFunctionData(
```
nalinbhardwaj, line 93: is there any advantage to doing a multi call given we don't take more than one path currently?

packages/daimo-api/src/api/getSwapRoute.ts
```diff
@@ -0,0 +1,139 @@
...
+    fromAcc: fromAccount,
+    receivedAt: t,
+    cacheUntil,
+    execDeadline,
+    toCoin: tokenOut,
+    routeFound: true,
+    toAmount: Number(amountOut),
+    execRouterAddress: UNISWAP_V3_02_ROUTER_ADDRESS,
+    execCallData: execCallData as Hex,
+    execValue: "0x00" as Hex,
```
nalinbhardwaj, line 113: should double check ETH swaps still work

packages/daimo-api/src/api/getSwapRoute.ts
kayleegeorge: true the multicall structure is odd -- the uniswap router always returns it and I was just trying to copy the format but ill try without right now and test

apps/daimo-clippy/src/handlers.ts
```diff
@@ -68,9 +68,9 @@ const commands: Record<string, Command> = {
     help: "Gets name, address and balance of a user. Args: [user = name or addr]",
     fn: getUser,
   },
-  "get-uniswap": {
-    help: "Gets the best Uniswap route for a given token to USDC. Args: [num=1.23, token=DAI]",
-    fn: getUniswapRoute,
+  "get-swap-quote": {
+    help: "Gets the best swap quote for fromToken to toToken. Args: [fromAmount=1.23, fromToken=DAI, toToken=USDC]",
+    fn: getSwapQuote,
```
dcposch, line 73: 🚀

apps/daimo-clippy/src/handlers.ts
```diff
@@ -146,29 +146,54 @@ async function getTokenList(): Promise<TokenList> {
...
-  const strToken = kwargs.get("token");
-  if (!strN || !strToken) return "Must specify num and token";
+// Gets a swap quote from the onchain contract.
+async function getSwapQuote(kwargs: Map<string, string>): Promise<string> {
+  const strN = kwargs.get("fromAmount");
+  const strFromToken = kwargs.get("fromToken");
+  const strToToken = kwargs.get("toToken");
+  const strDaimoChainId = kwargs.get("daimoChainId");
+  if (!strN || !strFromToken || !strToToken || !strDaimoChainId)
+    return "Must specify fromAmount, fromToken and toToken";
```
dcposch, line 156: dont need kwArg fro daimoChainId, can use the one from env

kayleegeorge: copy

packages/contract/src/DaimoUSDCSwapper.sol
```diff
@@ -0,0 +1,302 @@
...
+    IERC20 tokenA,
+    IERC20 tokenB
+  ) public view returns (address bestPool, int24 tick, uint24 bestFee) {
+    uint128 bestLiquidity = 0;
+    for (uint256 i = 0; i < oracleFeeTiers.length; i++) {
+      address pool = oraclePoolFactory.getPool({
+        tokenA: address(tokenA),
+        tokenB: address(tokenB),
+        fee: oracleFeeTiers[i]
+      });
```
dcposch, line 91: how do we handle symmetry? if there's a DAI-WETH pool and we're looking up (tokenA)WETH-(tokenB)DAI, will we still find it?

packages/contract/src/DaimoUSDCSwapper.sol
```diff
@@ -0,0 +1,302 @@
...
+      swapPath = abi.encodePacked(
+        address(tokenIn),
+        directFee,
+        address(tokenOut)
+      );
+    } else {
+      amountOut = hopAmountOut;
+      swapPath = swapPathHop;
+    }
+  }
```
dcposch, line 220: yesterday, we hotfixed API to pull `quoteDirect` instead of `quote` for a somewhat subtle reason.

our current quoter works as follows:
- uses the best liquidity pool *per pair* for a given path
- uses the highest price *per path* across (direct path + a few 2-pool paths)

problem is, what if the highest price is ~fake news? we saw an example yesterday.

if there is a hop path ~no liquidity (=best liquidity for at least one of the pairs along that route is very low) but highest price, we quote that path.

--

here's an idea for fixing this.
- `quote` > `quoteFromOracle` and just return price, no path. this is a rough estimate of best-available price, gas efficient, & robust to manipulation using TWAP/TWAL.

- a new function, `quoteBestPath`

use https://docs.uniswap.org/contracts/v3/reference/periphery/interfaces/IQuoterV2 to get the real amountOut for direct path vs best hop path and return that

dcposch (approved):
LGreatTM after comment

packages/contract/src/DaimoUSDCSwapper.sol
kayleegeorge: Ah good question --

kayleegeorge:
DO NOT MERGE !! BASESCAN DOING MAINTENANCE AND I WANT TO TEST FIRST

apps/daimo-clippy/src/handlers.ts
```diff
@@ -146,29 +147,54 @@ async function getTokenList(): Promise<TokenList> {
...
   return [
-    `Token: ${token.symbol} (${token.address})`,
-    `Best route for ${fromStr} to USDC: ${JSON.stringify(route, null, 2)}`,
+    `From ${fromToken.symbol} (${fromToken.address})`,
+    `To ${toToken.symbol} (${toToken.address})`,
+    `Fetched route for ${fromStr} to ${toToken.symbol}: ${JSON.stringify(
+      route,
+      null,
+      2
+    )}`,
```
dcposch, line 197: nit:
```ts
const routeStr = JSON.stringify(route, null, 2);
return [ ...
```

packages/contract/src/DaimoUSDCSwapper.sol
```diff
@@ -0,0 +1,322 @@
...
+
+      (
+        address poolOne,
+        ,
+        uint24 feeOne,
+        uint128 hopAmountOut
+      ) = getBestPoolTick(tokenIn, hopToken, amountIn);
+
+      if (poolOne == address(0)) continue;
+      if (hopAmountOut > _MAX_UINT128) continue;
```
kayleegeorge, line 181: arg redundant -- fixing and redeploying

packages/contract/src/DaimoUSDCSwapper.sol
dcposch: i think this will still cause issues.
* `getPool` accepts tokens in either order
* however `getPool(WETH, USDT)` and `getPool(USDT, WETH)` will return the same pool address
* as a result, both of those will end up with same value for `arithmeticMeanTick`!
i think this means one of those two cases will end up with an incorrect output amount

dcposch: nvm, looks good! `getQuoteAtTick` handles multiplying or dividing by the tick price depending on `address(tokenA) < address(tokenB)`

packages/contract/test/uniswap/Quoter.t.sol
```diff
@@ -0,0 +1,56 @@
...
+      _hopTokens: hopTokens,
+      _uniswapRouter: ISwapRouter(0x2626664c2603336E57B271c5C0b26F421741e481),
+      _oracleFeeTiers: oracleFeeTiers,
+      _oraclePeriod: 1 minutes,
+      _oraclePoolFactory: IUniswapV3Factory(
+        0x33128a8fC17869897dcE68Ed026d694621f6FDfD
+      )
+    });
+
+    assert(block.number == 15950101); // Block specific test
```
dcposch, line 38: 👍

dcposch (approved):
LGTM!

PR #1166, opened 2024-06-18
clippy: add "disable bonus" command

kayleegeorge:
Add a "disable bonus" for invite link feature for clippy. Invite link still remains valid, but the bonus is set to 0.

### DIFF
--- a/packages/daimo-api/src/db/db.ts
+++ b/packages/daimo-api/src/db/db.ts
@@ -416,13 +416,27 @@ export class DB {
 
   async updateInviteCode(row: UpdateInviteCodeArgs) {
     console.log(`[DB] updating invite code: ${JSON.stringify(row)}`);
+    let res;
+    if (row.maxUses) {
+      res = await this.pool.query<[], any[]>(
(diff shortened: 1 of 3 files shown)


nalinbhardwaj (approved):
lgtm

packages/daimo-api/src/db/db.ts
```diff
@@ -416,13 +416,27 @@ export class DB {
...
   async updateInviteCode(row: UpdateInviteCodeArgs) {
     console.log(`[DB] updating invite code: ${JSON.stringify(row)}`);
+    let res;
+    if (row.maxUses) {
+      res = await this.pool.query<[], any[]>(
+        `UPDATE invitecode SET max_uses = $1 WHERE code = $2`,
+        [row.maxUses, row.code]
+      );
+    }
+    if (row.bonusDollarInviter) {
```
dcposch, line 426: this won't run if we're setting `bonusDollarInviter` to 0, right?

recommend `if (row.bonusDollarInviter != null)` instead

(in general, JS "truthy"/"falsy" values have foot guns, so it's better to use explicit comparison most of the time)

PR #1167, opened 2024-06-18
upgrade relay bridge link

nalinbhardwaj:
they made a new page just for us


nalinbhardwaj:
PTAL post commit, tested new link with my account

PR #1168, opened 2024-06-19
shovel: eth_transfers, smaller batch, lower db load

dcposch:
cc @nalinbhardwaj 

testing in prod tonight, will revert if it causes issues


nalinbhardwaj:
Looks good, want to try with the `a7ed` hash shovel landed or skip that for now?

PR #1171, opened 2024-06-19
api: add currency: UAH

dcposch:


packages/daimo-common/src/currencies.ts
```diff
@@ -16,7 +16,8 @@ export const currencyRateUSD: CurrencyExchangeRate = {
 
 const data: [string, string, string, number][] = [
   ["Euro", "€", "EUR", 2],
-  ["Argentine Peso", "A$", "ARS", 0],
+  ["Argentine Peso", "ARS", "ARS", 0],
+  ["Ukrainian Hryvnia", "₴", "UAH", 0],
```
dcposch, line 20: these technically have 2 decimal points, but 1 UAH is worth about 2 cents so I don't think the decimals are useful

packages/daimo-common/src/currencies.ts
```diff
@@ -16,7 +16,8 @@ export const currencyRateUSD: CurrencyExchangeRate = {
 
 const data: [string, string, string, number][] = [
   ["Euro", "€", "EUR", 2],
-  ["Argentine Peso", "A$", "ARS", 0],
+  ["Argentine Peso", "ARS", "ARS", 0],
```
kayleegeorge, line 19: Did you mean to change the Argentinian Peso symbol to ARS or did you want to keep it A$ 

google says it can be both but A$ might save space in input row

kayleegeorge (approved):
comment otherwise LGTM

PR #1178, opened 2024-06-20
api: create token registry for known tokens

kayleegeorge:
Use a token registry instead of fetching foreign tokens in foreignCoinIndexer. 

Note: foreign token list does not include home coin inherently (so the foreignCoinIndexer doesn't pick up USDC transfers) so hav to add a flag to search for it when doing any coin send (sending from homeCoin).


nalinbhardwaj:
I think `getAddress` in viem returns a [checksummed address](https://docs.ethers.org/v5/api/utils/address/#:~:text=An%20Address%20is%20a%20DataHexString,or%20cut%20and%20paste%20issues) -- so we try to consistently use checksummed addresses everywhere.

Where did the all-lowercase address come from? I'd recommend changing that to the checksummed version using `getAddress` instead.

kayleegeorge:
Ah I see -- the coin gecko list by default has all-lowercase addresses. but updated now to reflect checksummed

packages/daimo-api/src/server/router.ts
```diff
@@ -224,18 +226,16 @@ export function createRouter(
...
         const { amountIn, fromToken, toToken, fromAccount, toAddr, chainId } =
           opts.input;
-        const foreignTokenList = foreignCoinIndexer.foreignTokens;
-        const route = await getSwapQuote({
+        return await getSwapQuote({
           amountInStr: amountIn,
-          tokenIn: fromToken,
-          tokenOut: toToken,
+          tokenIn: getAddress(fromToken),
+          tokenOut: getAddress(toToken),
```
dcposch, line 232: 👍

packages/daimo-api/src/server/tokenRegistry.ts
```diff
@@ -0,0 +1,94 @@
+import { ForeignCoin, USDbC, assert, getHomeCoinToken } from "@daimo/common";
+import { Address, getAddress, isAddress } from "viem";
+
+import { chainConfig } from "../env";
+import { fetchWithBackoff } from "../network/fetchWithBackoff";
+
+// Exclude native ETH
+export type ForeignToken = ForeignCoin & { token: Address };
+
+type TokenList = {
```
dcposch, line 10: `interface TokenList {`

packages/daimo-api/src/server/tokenRegistry.ts
```diff
@@ -0,0 +1,94 @@
+import { ForeignCoin, USDbC, assert, getHomeCoinToken } from "@daimo/common";
+import { Address, getAddress, isAddress } from "viem";
+
+import { chainConfig } from "../env";
+import { fetchWithBackoff } from "../network/fetchWithBackoff";
+
+// Exclude native ETH
+export type ForeignToken = ForeignCoin & { token: Address };
```
dcposch, line 8: `ForeignCoin` vs `ForeignToken` is going to be very easy to mix up. can we combine?

packages/daimo-api/src/server/tokenRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+    address: Address;
+    name: string;
+    symbol: string;
+    decimals: number;
+    logoURI?: string;
+  }[];
+  version: any;
+};
+
+// TODO: in the future, load a token registry for each chain.
```
dcposch, line 22: 👍 

we can also do a single TokenRegistry that knows where coins are on each chain, tbd

packages/daimo-api/src/server/tokenRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+          fullName: token.name,
+          symbol: token.symbol,
+          logoURI: largeLogo,
+        });
+      }
+    }
+
+    console.log(
+      `[TOKEN-REG] loaded ${this.foreignTokens.size} tokens on ${this.chainName}`
+    );
```
dcposch, line 66: 👍 looks clean

packages/daimo-common/src/foreignCoin.ts
```diff
@@ -66,3 +66,15 @@ export function isAmountDust(
 
   return true;
 }
+
+// Get home coin token for a given token address.
```
dcposch, line 70: not sure we need this function

what do you mean by "home coin token for a given token address"?

home coin is a property of a daimo account

the function behavior below is equivalent to just returning `daimoUSDC` in all cases, but with the address potentially overwritten to a different address? seems like a footgun

packages/daimo-api/src/server/tokenRegistry.ts
kayleegeorge: Yes sure -- this is what we had before, I think because "ETH" was separated out (token is "ETH" instead of the WETH address)

wasn't sure if there was a reason for that so left it in. but i can refactor

packages/daimo-api/src/api/getSwapRoute.ts
```diff
@@ -92,10 +98,13 @@ export async function getSwapQuote({
 
   if (!callData) return null;
 
-  if (!foreignTokenList) {
-    foreignTokenList = await fetchForeignTokenList();
+  let fromCoin;
+  // TODO: in future, check home coin token (for now, daimoUSDC)
+  if (tokenIn === daimoUSDC.address) {
+    fromCoin = daimoUSDC;
```
dcposch, line 104: 👍

packages/daimo-api/src/server/tokenRegistry.ts
```diff
@@ -0,0 +1,90 @@
+import { ForeignToken, USDbC, assert, getChainName } from "@daimo/common";
+import { Address, getAddress, isAddress, extractChain } from "viem";
+
```
dcposch, line 3: run Organize Imports > should fix automatically

packages/daimo-common/package.json
```diff
@@ -27,6 +27,7 @@
     "@noble/curves": "^1.2.0",
     "@scure/base": "^1.1.3",
     "@scure/bip39": "^1.3.0",
+    "@uniswap/sdk-core": "^4.2.1",
```
dcposch, line 30: this is a heavy dependency, & looks like we're using it just for the WETH addresses.

recommend hardcoding those (which we're already doing for DAI, USDbC etc)

minimizing dependencies important everywhere but esp for `daimo-common` since it gets imported everywhere else, including web and mobile

dcposch (approved):
LGreatTM after comments 🚢

packages/daimo-api/src/server/tokenRegistry.ts
dcposch: this will fix lint

Issue #1180, opened 2024-06-21
Optimize account history fetch


nounderline:
Some calls to `getAccountHistory` may take even to ~10s in some circumstances. The function itself is fetching data from:

- Internal indexers;
- Multiple Postgres queries through TRPC router or pg pool directly;
- external (!) RPC call through Viem.

Code, measurements: https://github.com/nounderline/240620-daimo-perf

nounderline:
Finalized block and exchange rates are bottlenecks:

```
┌───────────────────┬───────────────────────┬─────────────────────┬──────────────────────┬──────────────────────┐
│       event       │     min(duration)     │    max(duration)    │    avg(duration)     │ stddev_pop(duration) │
│      varchar      │        double         │       double        │        double        │        double        │
├───────────────────┼───────────────────────┼─────────────────────┼──────────────────────┼──────────────────────┤
│ linkedAccounts    │ 0.0019939999911002815 │ 0.12026500003412366 │ 0.003880708814813609 │ 0.007270018637333387 │
│ landlineSessions  │  0.003156999999191612 │ 0.12008500000229105 │ 0.006734735632555691 │  0.00980061516169723 │
│ namedAccounts     │  0.007663999916985631 │  0.6499700000276789 │ 0.016375045980368312 │  0.03993877100492821 │
│ chainGasConstants │  0.008114999975077808 │ 0.24601200000324752 │  0.01627666283518404 │ 0.020089551472207094 │
│ exchangeRates     │    1.4265860000159591 │   51.36471799999708 │    31.92754198467884 │    17.77380790597216 │
│ finBlock          │    26.507030999986455 │  20343.944205000007 │    703.9794512758592 │     3177.44886501936 │
│ balance           │  0.010079000028781593 │ 0.14357899996684864 │ 0.017577111110861958 │ 0.012388204015386468 │
│ filterTransfers   │   0.12827999994624406 │   2.796078000043053 │  0.20850022222229656 │  0.17386769688897366 │
└───────────────────┴───────────────────────┴─────────────────────┴──────────────────────┴──────────────────────┘
```

PR #1191, opened 2024-06-24
Streamline latest block access & finalized block caching

nounderline:
- Cache finalized block fetching globally. Before `getBlock` RPC call was made for every `getAccountHistory` request. Now the value is refreshed at most once every 1s. More: #1180
- Fix a bug where `AccountHistoryResult` could be one block behind. Now we get latest indexed block from Shovel database instead from internal `Watcher` indexer in `getAccountHistory`. This bug occurred when event loop called a new block notification callback in `Watcher.watch` before one in `onAccountUpdate` TRPC subscription.

### DIFF
--- a/packages/daimo-api/src/server/router.ts
+++ b/packages/daimo-api/src/server/router.ts
@@ -355,10 +355,8 @@ export function createRouter(
           address,
           inviteCode,
           sinceBlockNum,
-          watcher,
           vc,
           homeCoinIndexer,
-          ethIndexer,
           foreignCoinIndexer,
           profileCache,
           noteIndexer,
@@ -369,7 +367,8 @@ export function createRouter(
           keyReg,
           paymaster,
           db,
-          extApiCache
+          extApiCache,
+          await watcher.getShovelLatest()
         );
       }),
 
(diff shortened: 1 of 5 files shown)


packages/daimo-api/src/server/router.ts
```diff
@@ -756,42 +756,42 @@ export function createRouter(
...
-
-              emit.next(history);
-
-              lastEmittedBlock = history.lastBlock;
-            });
+              extApiCache,
+              blockNumber
+            );
+
+            if (lastEmittedBlock > history.lastBlock) {
```
dcposch, line 763: should this be `>=` ?

packages/daimo-api/src/utils/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * `cacheTime` is duration in ms after cached value is considred expired,
+ * and must be refreshed before it can be returned.
+ *
+ * `staleTime` is an optional duration that refreshes
+ * the cache before it expires. Use it to avoid blocking after cache expires.
+ * Must be smaller than `cacheTime`.
+ *
+ */
+export function lazyCache<T>(
```
dcposch, line 14: this is interesting. feels like a lot of complexity for just `getFinalizedBlock`

finalizedBlock changes slowly & doesn't affect UI latency. is the goal to reduce JSON RPC API calls?

dcposch (approved):
Left comments, LGTM otherwise

packages/daimo-api/src/utils/cache.ts
nounderline: Finalized block is done on every account history fetch. It was [second slowest operation](https://github.com/daimo-eth/daimo/issues/1180#issuecomment-2186485047) in `getAccountHistory`. Now we share finalized block data across all users limiting RPC calls from variable `n` to constant `2` every ~2 sec, where `n` is number of connected users.

PR #1193, opened 2024-06-25
contract: DaimoFastCCTP

dcposch:


packages/contract/src/DaimoFastCCTP.sol
```diff
@@ -0,0 +1,241 @@
...
+  function destroy() public onlyCreator {
+    // This use of SELFDESTRUCT is compatible with EIP-6780.
+    // EphemeralHandoffs are deployed, then destroyed in the same transaction.
+    // solhint-disable-next-line
+    selfdestruct(_creator);
+  }
+
+  // This list only restricts destination chains = compatiable DaimoAccountV2
+  // home chains. The source chain could be any chain CCTP adds in the future.
+  function _chainIDToDomain(uint256 chainID) private pure returns (uint32) {
```
kayleegeorge, line 232: nice

packages/contract/src/DaimoFastCCTP.sol
```diff
@@ -0,0 +1,241 @@
...
+      fromChainID,
+      fromAddr,
+      fromAmount,
+      toChainID,
+      toAddr,
+      toAmount,
+      toToken,
+      nonce
+    );
+    handoff.destroy();
```
nalinbhardwaj, line 95: just to confirm, you only need the address of the handoff here right? might be cleaner to make a util function for computing the create2 address or adding a comment

packages/contract/src/DaimoFastCCTP.sol
```diff
@@ -0,0 +1,241 @@
...
+
+    // LP pays original recipient
+    toToken.safeTransferFrom(msg.sender, toAddr, toAmount);
+  }
+
+  // Claims a CCTP transfer to its current recipient. If fastFinish... was
+  // called for this transfer, then recipient = the LP who fronted the amount.
+  // Otherwise, the recipient remains the original toAddr, eg Bob.
+  function claimTransfer(
+    ICCTPReceiver cctpReceiver,
```
nalinbhardwaj, line 114: hmm I think you might want to hardcode the cctpReceiver address as a constructor arg (and use create3 to deploy it?)

otherwise I can see some vulnerable edge case around being able to create a fake CCTP receiver that accepts any message/signature and calling this function with those (but not sure)

packages/contract/src/DaimoFastCCTP.sol
nalinbhardwaj: should this mapping be a constructor arg?

kayleegeorge: yea the daimoCCTPBridger currently takes in inputChainIds and outputDomains as constructor args --> maybe that would be better

PR #1199, opened 2024-06-26
common: add chain and token information

kayleegeorge:
Add chain and foreign token constants for account upgrade. 

Fix: Prevent send any-stable on testnet (no swapper / Uniswap pools).

### DIFF
--- a/packages/daimo-common/src/chain.ts
+++ b/packages/daimo-common/src/chain.ts
@@ -9,13 +9,23 @@ import { DaimoChain } from "@daimo/contract";
 
 import {
   ForeignToken,
+  arbitrumSepoliaUSDC,
   arbitrumSepoliaWETH,
+  arbitrumUSDC,
   arbitrumWETH,
+  avalancheFujiUSDC,
+  avalancheUSDC,
   avalancheWETH,
+  baseSepoliaUSDC,
   baseSepoliaWETH,
+  baseUSDC,
   baseWETH,
+  optimismSepoliaUSDC,
   optimismSepoliaWETH,
+  optimismUSDC,
   optimismWETH,
+  polygonSepoliaUSDC,
+  polygonUSDC,
   polygonWETH,
 } from "./foreignToken";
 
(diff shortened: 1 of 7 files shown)


packages/daimo-common/src/chain.ts
```diff
@@ -1,31 +1,178 @@
...
+  baseUSDC,
+  baseWETH,
+  optimismSepoliaUSDC,
+  optimismSepoliaWETH,
+  optimismUSDC,
+  optimismWETH,
+  polygonSepoliaUSDC,
+  polygonUSDC,
+  polygonWETH,
+} from "./foreignToken";
```
dcposch, line 30: 🚀

packages/daimo-common/src/chain.ts
```diff
@@ -1,31 +1,178 @@
...
 
 export type AccountChain = {
   chainId: number;
   name: string;
-  nativeWETH: ForeignToken;
+  CCTPDomain: number;
```
dcposch, line 35: `cctpDomain`

packages/daimo-common/src/chain.ts
```diff
@@ -1,31 +1,178 @@
...
   name: "baseSepolia",
+  CCTPDomain: 6,
+  bridgeCoin: baseSepoliaUSDC,
   nativeWETH: baseSepoliaWETH,
+  isTestnet: true,
+};
+
+export const Arbitrum: AccountChain = {
+  chainId: 42161,
+  name: "Arbitrum",
```
dcposch, line 60: case > make consistent with `base`/`baseSepolia` above

packages/daimo-common/src/foreignToken.ts
```diff
@@ -13,6 +35,15 @@ export type ForeignToken = {
   logoURI?: string;
 };
 
+export enum TokenLogo {
+  ETH = "https://assets.coingecko.com/coins/images/279/large/ethereum.png",
+  USDC = "https://assets.coingecko.com/coins/images/6319/large/usdc.png",
+  USDT = "https://assets.coingecko.com/coins/images/325/large/tether.png?1547034089",
+  DAI = "https://assets.coingecko.com/coins/images/9956/large/dai-multi-collateral-mcd.png?1574218774",
+}
```
dcposch, line 43: 👍 

doesn't have to be in this PR but in future we can inline these as small base64 images

dcposch (approved):
Nice cleanup!

packages/daimo-common/src/chain.ts
kayleegeorge: good catch

PR #1204, opened 2024-07-02
clippy: fix typo

kayleegeorge:
typo in UpdateInviteCodeArgs: `bonusDollarInviter` instead of `bonusDollarsInviter` --> causing invite link args to not update properly


dcposch (approved):
LGTM. Why didn't this have a type error before?

PR #1217, opened 2024-07-19
mobile: Deposit from Lemon

dcposch:
- [x] ~Verify Lemon supports Base~
- [x] Use Polygon & PR relies on DAv2


dcposch:
Will revisit after #1224

PR #1218, opened 2024-07-19
mobile: revert to normal polling, fix foreignCoinIndexer

dcposch:
### Mobile > API polling

websocket implementation was unreliable. fixes spurious offline indicator.

we are still using websockets to talk to ethererum RPC (both API & shovel), & postgres NOTIFY for api to get rapid updates from shovel.

### foreignCoinIndexer

This PR deletes the materialized views that were causing us grief. As a stopgap till a more permanent indexer fix, this should restore inbox functionality, including receiving ETH.


kayleegeorge (approved):
nice cleanup!!

PR #1224, opened 2024-07-22
mobile: support DaimoAccountV2

dcposch:
- [x] ShovelSquared: update `daimo_transfers` table from Jumbo DB = large shovel for all transfers on all chains
- [x] Deploy new DaimoAccountV2
- [x] API handles both DAv1 and V2.
    - [ ] Distinguish DAv1 from V2 onchain. Currently implemented in a stopgap way with a timestamp cutover. Implementing this robustly requires either 1. indexing the ERC1967 `Upgraded` event or 2. loading & DB caching the current implementation address for each encountered  DaimoAccount
- [ ] ShovelSquared: ingest `daimo_transfers` from foreign chains as well

+ automatic any-chain receive should work after this branch by pointing `swapbot` to the `daimo_transfers` table ⚡️


dcposch:
Can revisit after hack week.

PR #1228, opened 2024-07-23
common: add utils to common 

kayleegeorge:
- move `retryBackoff` from `daimo/api` to `@daimo/common` so it can be used in swapbot
- Add some constants to `chain.ts` and `foreignToken.ts`


packages/daimo-common/src/chain.ts
```diff
@@ -24,9 +25,11 @@ import {
   optimismSepoliaWETH,
   optimismUSDC,
   optimismWETH,
-  polygonSepoliaUSDC,
+  polygonAmoyUSDC,
   polygonUSDC,
   polygonWETH,
+  ethereumSepoliaUSDC,
+  ethereumUSDC,
```
dcposch, line 32: 👍

dcposch (approved):

PR #1235, opened 2024-07-26
i18n: polish

kayleegeorge:
- rm codegen i18n type; just use `en` as canonical type using `LanguageDefinition`


dcposch (approved):
Looks great.
can fix lint by adding an ignore comment or using LanguageDefinition type in one other place

PR #1238, opened 2024-07-29
api: switch to IndexDB

dcposch:
Decoupling `daimo-api` from indexer implementation details, using #1239


packages/daimo-api/src/contract/SwapClogMatcher.ts
```diff
@@ -20,39 +23,56 @@ export class SwapClogMatcher {
...
         blockHash: bytesToHex(row.block_hash, { size: 32 }),
         blockNumber: BigInt(row.block_num),
         transactionHash: bytesToHex(row.tx_hash, { size: 32 }),
-        transactionIndex: row.tx_idx,
-        logIndex: row.log_idx,
-        address: getAddress(bytesToHex(row.log_addr, { size: 20 })),
+        transactionIndex: Number(row.tx_idx),
+        logIndex: Number(row.sort_idx),
+        address:
+          row.token == null
```
andrewliu08, line 68: in s^2, should we fill the token to zero address instead of having it as null?

dcposch: good call, S^2 does that now

dcposch:
image

andrewliu08:
should we also test remove and create account?

andrewliu08 (commented):
lgtm if everything was tested

dcposch:
* Tested add device + remove device
* V1>V2 upgrade over in this PR: #1224

PR #1243, opened 2024-08-03
web: I18N

dcposch:
cc @Ethanol48 

I've implemented a minimalist I18N setup that matches the approach we took for `daimo-api`, `mobile`, and `common`.

- All `page.tsx` components will be Next server components. They can handle metadata, data fetching, and locale. See the two `page.tsx` modified in this PR for examples.
- All actual rendering can happen in client components split out from the parent page. See, for example, `LinkPage.tsx`. These take `lang` as a prop, and they use `I18NProvider` to provide language context for child components.
- Finally, the interface components use the `useI18N()` hook to get internationalization.

The legacy `/link/...` page doesn't need to be translated. I've marked it as deprecated.

Have a great weekend & let me know if you run into any issues 🫡

### DIFF
--- a/apps/daimo-web/src/app/HomePage.tsx
+++ b/apps/daimo-web/src/app/HomePage.tsx
@@ -1,0 +1,25 @@
+"use client";
+
+import { SectionFAQ } from "./SectionFAQ";
+import { SectionHero } from "./SectionHero";
+import { SectionTestimonial } from "./SectionTestimonial";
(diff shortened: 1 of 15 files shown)


Ethanol48:
I have a small problem, there's some fields that needs to be in i18n but I don't think that can't be use, like in layout.tsx, I'm not familiar with Context in React, how can I access i18n in the layout?

dcposch:
Superseded by #1244

Issue #1246, opened 2024-08-06
Incident: unable to send due to nonce miscalculation


dcposch:
## Summary

New user unable to create account + all sends failing for ~30 minutes due to incorrect nonces.

## Timeline

2024-08-06 Argentina time:
- 11:46am: last successful transaction, [0x5584 create account](https://basescan.org/tx/0x1c9a34975ea5c61154ce13ce69edd7fd81a2472ef21891ff0645660ea7c05584)
- 11:47am. First failure.
- 12:20pm. Restart API server
- 12:27pm: [first successful transaction](https://basescan.org/tx/0xd6677f76e4ad866ee43650c4e8587b709a761648e8e304b40b4acdb51b277210) post outage

#### No new successful transactions, incrementing nonce

![image](https://github.com/user-attachments/assets/3be921d9-f0cc-4878-88e7-ba3d63880cc5)

#### Starts with "transaction underpriced" @ nonce 23064

![image](https://github.com/user-attachments/assets/3bc3bc23-3728-47bb-af5c-d59fc99e98d8)

![image](https://github.com/user-attachments/assets/276cc18a-b24b-4e2c-84cd-aaa5d0aaa17b)

## Fix

- [ ] **More robust nonce calculation.** Ensure concurrent ops don't try the same nonce (= what we're currently attempting via `nextNonce++`), but also ensure nonce doesn't diverge from onchain transaction count indefinitely.

We can also think about a strategy for underpriced transactions. We can set our maxBaseFee higher, within reason, but this will still be exceeded occasionally. Retry when gas is lower, or just show user a friendly error?

PR #1247, opened 2024-08-07
API: fix nonce issue

andrewliu08:
Fixes nonce issue in https://github.com/daimo-eth/daimo/issues/1246

## Example
5 transactions with nonces 1, 2, 3, 4, 5 get sent. Tx 1 gets rejected but tx 2-5 are queued by the sequencer.

Without the fix, the client would send new txs with nonces 6, 7, 8, ... These new txs will fail with a "receipt error" because a tx with nonce 1 hasn't been sent yet.

With the fix, the client will time out waiting for the receipts of txs 1-5, then it would send the next tx with nonce 1. Txs 2-5 will get processed. `updateNonce` will query the next nonce from the RPC and set the nonce of the next tx to 6.

### DIFF
--- a/packages/daimo-api/src/network/viemClient.ts
+++ b/packages/daimo-api/src/network/viemClient.ts
@@ -190,30 +190,14 @@ export class ViemClient {
     );
   }
 
-  private async waitForReceipt(hash: Hex, maxWaitMs: number = 10_000) {
+  private async waitForReceipt(hash: Hex, timeoutMs: number = 10_000) {
     let receiptStatus = "";
 
     try {
-      const receiptPromise = this.publicClient.waitForTransactionReceipt({
+      const receipt = await this.publicClient.waitForTransactionReceipt({
         hash,
+        timeout: timeoutMs,
       });
-      const timeoutPromise = new Promise((_, reject) =>
-        setTimeout(
-          () =>
-            reject(
-              new Error(
-                `Timed out after ${maxWaitMs}ms waiting for receipt ${hash}`
-              )
-            ),
-          maxWaitMs
-        )
-      );
-
-      // Wait at most maxWaitMs milliseconds for the receipt
-      const receipt = (await Promise.race([
-        receiptPromise,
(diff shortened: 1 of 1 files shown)


packages/daimo-api/src/network/viemClient.ts
```diff
@@ -188,18 +190,50 @@ export class ViemClient {
     );
   }
 
-  private async waitForReceipt(hash: Hex) {
+  private async waitForReceipt(hash: Hex, maxWaitMs: number = 10_000) {
+    let receiptStatus = "";
+
     try {
-      const receipt = await this.publicClient.waitForTransactionReceipt({
+      const receiptPromise = this.publicClient.waitForTransactionReceipt({
```
dcposch, line 197: can pass `timeout`

packages/daimo-api/src/network/viemClient.ts
```diff
@@ -188,18 +190,50 @@ export class ViemClient {
...
+      await this.lockNonce.acquireAsync();
+
+      if (receiptStatus === "success") {
+        this.nextNonce += 1;
+        this.pendingTxCount -= 1;
+      } else {
+        this.pendingTxCount -= 1;
+      }
+
+      this.lockNonce.release();
```
dcposch, line 236: we might not need lockNonce at all (here or other spot)

packages/daimo-api/src/network/viemClient.ts
```diff
@@ -188,18 +190,50 @@ export class ViemClient {
...
     }
   }
```
dcposch, line 239: in updateNonce below, we no longer want `blockTag: "pending"`

packages/daimo-api/src/network/viemClient.ts
andrewliu08: will leave this in for now and check what `AwaitLock` is actually doing

dcposch (approved):
LGreatTM

PR #1248, opened 2024-08-09
mobile: add any chain UI

kayleegeorge:
Screenshot 2024-08-08 at 10 52 52 PM
Screenshot 2024-08-08 at 10 52 41 PM
Screenshot 2024-08-08 at 11 32 16 PM


apps/daimo-mobile/src/view/screen/send/CoinDisplay.tsx
```diff
@@ -1,113 +1,211 @@
...
+    case optimismSepolia:
+      return IconOptimism;
+    case polygon:
+    case polygonAmoy:
+      return IconPolygon;
+    case arbitrum:
+    case arbitrumSepolia:
+      return IconArbitrum;
+  }
+};
```
andrewliu08, line 57: maybe move this to daimo common in the future

kayleegeorge: Reason is loading from mobile bundle (../../../../assets/...) vs. loading from fetched source url 

I think the former is faster? but wdyt

andrewliu08 (approved):
lgtm

PR #1264, opened 2024-08-12
Revert "API: fix nonce issue"

andrewliu08:
Reverts daimo-eth/daimo#1247

### DIFF
--- a/packages/daimo-api/src/network/viemClient.ts
+++ b/packages/daimo-api/src/network/viemClient.ts
@@ -120,7 +120,6 @@ export class ViemClient {
   // Lock to ensure sequential nonce for walletClient writes
   private lockNonce = new AwaitLock();
   private nextNonce = 0;
-  private pendingTxCount = 0;
   public account: Account;
 
   constructor(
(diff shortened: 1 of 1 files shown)


dcposch:
LGTM. 

Follow-ups:
- We should reproduce the problem on staging.
- **Updated fix. After this PR, the original issue (gas spike > tx rejected > nonce skipped) is a possibility again.**

dcposch (approved):

packages/daimo-api/src/network/viemClient.ts
```diff
@@ -258,17 +241,16 @@ export class ViemClient {
         `[VIEM] tx ${localTxId} ${elapsedMs()}ms: got nonce ${this.nextNonce}`
       );
 
-      // Override nonce with nonce estimate. Optimistically assumes that
-      // all pending transactions will succeed.
-      args.nonce = this.nextNonce + this.pendingTxCount;
```
dcposch, line 263: this line was missing an important log. see log above ("...got nonce: ") logging an incorrect nonce, vs the one we're submitting, and not logging `pendingTxCount`.

PR #1267, opened 2024-08-12
mobile: show invite link on QR page

dcposch:
Show our invite link, if we have one. Otherwise, just link to our account.
Either way, if a user w/ app installed scans, they go to our account page.


dcposch:
Can revisit

PR #1272, opened 2024-08-14
Fix misc typos

ottok:
Includes writing 'thru' as 'through'.

All changes are non-functional in comments or READMEs and should have no effect on the app, just helps maintain the code base.

This is easiest to review with command `git show --word-diff` if pulling locally.


dcposch (approved):
Nice

Curious, how did you find us? Do you have a script for these types of fixes?

PR #1273, opened 2024-08-14
Clippy: Add commands for creating tags

andrewliu08:


apps/daimo-clippy/src/handlers.ts
```diff
@@ -77,6 +79,18 @@ const commands: Record<string, Command> = {
...
+    fn: getTag,
+  },
+  "create-tag": {
+    help: "Create a new tag. Associates https://daimo.com/l/t/<tag> with a link. Args: [tag, link]",
+    fn: createTag,
+  },
+  "update-tag": {
+    help: "Update a tag. Associates https://daimo.com/l/t/<tag> with a new link. Args: [tag, link]",
+    fn: updateTag,
+  },
```
dcposch, line 93: how about just a single one , `tag`?

```
@Clippy tag t=abc url=https://google.com
```

dcposch: if already exists, can return an error "abc already exists", then user can run

```
@Clippy tag t=abc new-url=https://google.com
```

apps/daimo-clippy/src/handlers.ts
```diff
@@ -15,6 +15,8 @@ import { getJSONblock, parseKwargs, unfurlLink } from "./utils";
 
 const apiKey = assertNotNull(process.env.DAIMO_API_KEY);
 
+const clippyTagUpdateToken = "clippy-tag-update-token";
```
dcposch, line 18: env var?

packages/daimo-api/src/db/db.ts
```diff
@@ -235,6 +235,22 @@ export class DB {
     }));
   }
 
+  async createTagRedirect(tag: string, link: string, updateToken: string) {
```
dcposch, line 238: could add `updateToken` as optional arg to saveTag..

dcposch (approved):
LGTM otherwise 🚀

PR #1276, opened 2024-08-16
Mobile: handle deeplinking for tags

andrewliu08:

PR #1278, opened 2024-08-19
mobile: handle tag deeplinks during onboarding

andrewliu08:
When clicking a tag deeplink that redirects to an invite link, it doesn't autofill the invite code.

Messed up and didn't notice this issue in this PR :(
https://github.com/daimo-eth/daimo/pull/1276

## Testing
Created an invite code called `test` and a tag called `test` which redirects to `https://daimo.com/l/invite/test`.
Created an invalid invite code called `test` and a tag called `test1`.
Created a tag called `request` which redirects to `https://daimo.com/l/request?to=vitalik.eth&c=84532&t=usdc`

| link       | result       |
|----------------|----------------|
| daimo://t/test | redirects to "choose username"  |
| daimo://invite/test  | redirects to "choose username"  |
| daimo://t/test1  | redirects to "enter invite code" |
| daimo://invite/test1  | redirects to "enter invite code" |
| daimo://t/request  | redirects to "enter invite code" |
| daimo://request?to=vitalik.eth&c=84532&t=usdc  | redirects to "enter invite code" |

### DIFF
--- a/apps/daimo-mobile/src/common/nav.ts
+++ b/apps/daimo-mobile/src/common/nav.ts
@@ -1,5 +1,6 @@
 import {
   DAv2Chain,
+  DaimoInviteCodeStatus,
   DaimoLink,
   DaimoLinkAccount,
   DaimoLinkInviteCode,
@@ -13,7 +14,6 @@ import {
   TransferClog,
   getEAccountStr,
   parseDaimoLink,
-  parseInviteCodeOrLink,
 } from "@daimo/common";
 import { DaimoChain, daimoChainFromId } from "@daimo/contract";
 import { NavigatorScreenParams, useNavigation } from "@react-navigation/native";
@@ -34,7 +34,7 @@ import {
   getInitialDeepLink,
   markInitialDeepLinkHandled,
 } from "../logic/deeplink";
-import { fetchInviteLinkStatus, fetchLinkStatus } from "../logic/linkStatus";
+import { fetchLinkStatus } from "../logic/linkStatus";
 import { MoneyEntry } from "../logic/moneyEntry";
 import { Account } from "../storage/account";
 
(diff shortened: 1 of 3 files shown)


apps/daimo-mobile/src/common/nav.ts
```diff
@@ -214,13 +214,44 @@ export async function handleOnboardingDeepLink(
...
+        linkStatus.link
+      )} is not an invite link`
+    );
+    nav.navigate("CreateNew");
+    return;
+  }
+
+  const inviteStatus = linkStatus as DaimoInviteCodeStatus;
   const isAndroid = Platform.OS === "android";
-  if (inviteLink && (await fetchInviteLinkStatus(dc, inviteLink))?.isValid) {
```
dcposch, line 220: i like the simplicity of the previous way here.

also, i think we can fix this server-side too!

fetchInviteLinkStatus does two things:
- call `fetchLinkStatus` via RPC
- send the result thru `getInviteStatus` locally

...we can fix `fetchLinkStatus` so that, when given a tag URL, it returns

```
{
   link: { /* the resolved link */ },
   ...
}
```

andrewliu08: hmm... would be good to move more of the logic server-side.

however, the `fetchInviteLinkStatus` code isn't being called because this line parses tags as `undefined`, so i think we need to do an app release either way
```
const inviteLink = parseInviteCodeOrLink(str);
```

dcposch: makes sense.
yeah, i think this section in parseInviteCodeOrLink could be simplified

before:
```ts
const link = parseDaimoLink(str);
  if (link != null && ["invite", "requestv2", "notev2"].includes(link.type)) {
    return link;
  }

return undefined;
```
  
after:

```ts
return parseDaimoLink(str);
```

PR #1282, opened 2024-08-20
api: correct proposed swaps

dcposch:
- load from daimo_transfers in orders
- correct proposedSwap timestamps
- better logging

### DIFF
--- a/packages/daimo-api/src/contract/foreignCoinIndexer.ts
+++ b/packages/daimo-api/src/contract/foreignCoinIndexer.ts
@@ -78,6 +78,8 @@ export class ForeignCoinIndexer extends Indexer {
           ])
           .where("chain_id", "=", chainConfig.chainL2.id)
           .where((eb) => eb.between("block_num", "" + from, "" + to))
+          .orderBy("block_num", "asc")
+          .orderBy("sort_idx", "asc")
           .execute()
     );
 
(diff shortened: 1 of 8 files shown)


packages/daimo-api/src/contract/foreignCoinIndexer.ts
```diff
@@ -78,6 +79,8 @@ export class ForeignCoinIndexer extends Indexer {
           ])
           .where("chain_id", "=", chainConfig.chainL2.id)
           .where((eb) => eb.between("block_num", "" + from, "" + to))
+          .orderBy("block_num", "asc")
+          .orderBy("sort_idx", "asc")
```
andrewliu08, line 83: does this use the index on the table? or do we need to do
```
.orderBy("block_num", "asc")
.orderBy("tx_idx", "asc")
.orderBy("sort_idx", "asc")

File: packages/daimo-api/src/contract/homeCoinIndexer.ts [hunk 2]
```diff
@@ -84,6 +84,8 @@ export class HomeCoinIndexer extends Indexer {
           .where("chain_id", "=", chainConfig.chainL2.id)
           .where((e) => e.between("block_num", "" + from, "" + to))
           .where("token", "=", hexToBuffer(chainConfig.tokenAddress))
+          .orderBy("block_num", "asc")
+          .orderBy("sort_idx", "asc")
```
andrewliu08, line 88: same here

andrewliu08 (approved):
lgtm

packages/daimo-api/src/contract/foreignCoinIndexer.ts
dcposch: order by is nbd (=sorting a small result set), important part is that the WHERE clauses are a prefix of the index

Issue #1275, opened 2024-08-15
Download link component in phone browser doesn't reflect i18n


Ethanol48:
i18n needs to be implemented for mobile view

apps/daimo-web/src/components/DownloadLink.tsx

dcposch:
"phone explorer" = mobile browser?

Ethanol48:
I created the PR I'm going to close the issue.

Ethanol48:
Problem solved by PR

PR #1284, opened 2024-08-20
web: i18n for download links

Ethanol48:
Created a getter function for  to reflect i18n and small change download page to add a word to i18n

### DIFF
--- a/apps/daimo-web/src/app/download/other/page.tsx
+++ b/apps/daimo-web/src/app/download/other/page.tsx
@@ -4,11 +4,12 @@ import React from "react";
 
 import { PrimaryButton } from "../../../components/buttons";
 import { useI18N } from "../../../i18n/context";
-import { downloadMetadata } from "../../../utils/platform";
+import { getDownloadMetadata } from "../../../utils/platform";
 
 export default function DownloadPage() {
   const i18n = useI18N();
   const i18 = i18n.download;
+  const downloadMetadata = getDownloadMetadata(i18n);
 
   return (
     <main className="max-w-2xl mx-auto px-4">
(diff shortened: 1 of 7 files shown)


Ethanol48:
Fixed behavior mentioned in #1275

apps/daimo-web/src/app/download/other/page.tsx
```diff
@@ -57,7 +58,7 @@ export default function DownloadPage() {
           </div>
           <div className="flex flex-row w-1/6 justify-center">
             <div className="h-[350px] flex flex-row items-center justify-center">
-              <p className="text-xl font-semibold text-grayMid">OR</p>
+              <p className="text-xl font-semibold text-grayMid">{i18.OR()}</p>
```
dcposch, line 61: 👍

apps/daimo-web/src/utils/platform.ts
```diff
@@ -12,6 +13,7 @@ export function detectPlatform(ua: string): PlatformType {
   return "other";
 }
 
+// necesary for some routes
```
dcposch, line 16: I think we can delete downloadMetadata > use `getDownloadMetadata` everywhere

dcposch (approved):
Otherwise LGreatTM

apps/daimo-web/src/utils/platform.ts
Ethanol48: I'm not sure how to do that at the moment, there's this file `route.tsx` in the path /download that detectes the platform, I don't think that the hook `useI18n()` can be used, and I didn't want to over complicate the code

Ethanol48:
I made the change to only use the function

PR #1208, opened 2024-07-11
Deposit from Landline

andrewliu08:

PR #1289, opened 2024-08-21
Deposit from Landline

andrewliu08:
A newer version of this PR with a rebase on master
https://github.com/daimo-eth/daimo/pull/1208

1. created deposit page with slider between withdraw/deposit
2. added logic for calling the deposit function on the landline API 
3. cleaned up API interface between landline and daimo API
4. created feature flag class instead of using env var
5. generate landline session url server-side instead of in the mobile app
6. create "landlineDeposit" offchain action to authenticate deposits


apps/daimo-mobile/src/action/useLandlineDeposit.ts
```diff
@@ -0,0 +1,49 @@
...
+import { ActHandle, useActStatus } from "../action/actStatus";
+import { getRpcFunc } from "../logic/trpc";
+import { Account } from "../storage/account";
+
+interface UseLandlineDepositArgs {
+  account: Account;
+  recipient: { landlineAccountUuid: string };
+  dollarsStr: string;
+  memo?: string;
+}
```
dcposch, line 14: 🔥

apps/daimo-mobile/src/action/useLandlineDeposit.ts
```diff
@@ -0,0 +1,49 @@
...
+  recipient,
+  dollarsStr,
+  memo,
+}: UseLandlineDepositArgs): ActHandle & { exec: () => Promise<void> } {
+  const [as, setAS] = useActStatus("useLandlineDeposit");
+
+  const exec = useCallback(async () => {
+    setAS("loading", "Creating deposit");
+    try {
+      // TODO: authenticate call to the deposit RPC function
```
dcposch, line 27: this is the part where we need session keys

dcposch: alternative: we can use an ERC-1271 signed message, just like we do in the Link Farcaster flow.

will require FaceID, but that's a feature IMO: feels correct to authenticate equally for (deposit from bank) vs (withdraw to bank)

apps/daimo-mobile/src/i18n/languages/en.ts
```diff
@@ -454,6 +454,10 @@ export const en = {
   contactDisplay: {
     requestedBy: () => `Requested by`,
   },
+  // LandlineDepositButton.tsx
+  landlineDepositButton: {
+    holdButton: () => "HOLD TO DEPOSIT",
```
dcposch, line 459: 🔥

apps/daimo-mobile/src/action/useLandlineDeposit.ts
andrewliu08: should we ship a version with deposits disabled and then create a follow-up PR with authentication for deposits?

apps/daimo-mobile/src/storage/storedTypes.ts
```diff
@@ -241,12 +241,14 @@ interface StoredV16ForeignCoin {
...
   accountName: string;
-  lastFour: string;
+  accountNumberLastFour: string;
+  bankCurrency: string;
   liquidationAddress: Address;
-  chain: string;
-  destinationCurrency: string;
-  bankLogo?: string;
+  liquidationChain: string;
+  liquidationCurrency: string;
```
dcposch, line 252: hmm, adding non-optional fields to an existing StoredModel is generally bad & can cause crashes.

It should be OK in this case since we haven't shipped Landline > nobody except us should have these in their stored account.

Can you confirm that nobody outside Daimo could have one of these yet?

apps/daimo-mobile/src/view/screen/DepositScreen.tsx
```diff
@@ -97,8 +102,7 @@ function LandlineConnect() {
     <LandlineOptionRow
       cta={i18.landline.cta()}
       title={i18.landline.title()}
-      // TODO(andrew): Update with real landline logo
-      logo={IntroIconEverywhere}
+      logo={LandlineLogo}
```
dcposch, line 96: 👍

apps/daimo-mobile/src/view/screen/send/SendTransferButton.tsx
```diff
@@ -160,7 +160,7 @@ export function SendTransferButton({
       case "error":
         return (
           <LongPressBigButton
-            title={i18.holdButton()}
+            title="HOLD TO SEND"
```
dcposch, line 163: ?

packages/daimo-api/src/api/featureFlag.ts
```diff
@@ -0,0 +1,9 @@
+import { EAccount } from "@daimo/common";
+
+export class FeatFlag {
+  public static landline(account: EAccount & { name: string }) {
+    return ["dcposch", "klee", "andrewliu", "nibnalin", "hanna"].includes(
+      account.name
```
dcposch, line 6: 👍

packages/daimo-api/src/server/router.ts
```diff
@@ -684,6 +686,29 @@ export function createRouter(
...
+        z.object({
+          daimoAddress: zAddress,
+          landlineAccountUuid: z.string(),
+          amount: zDollarStr,
+          memo: z.string().optional(),
+        })
+      )
+      .mutation(async (opts) => {
+        // TODO: add authentication this endpoint
+        const { daimoAddress, landlineAccountUuid, amount, memo } = opts.input;
```
dcposch, line 700: could add `, signature`, see `updateProfileLinks`

dcposch (approved):
🔥 
LGreatTM after comments

apps/daimo-mobile/src/storage/storedTypes.ts
andrewliu08: the landline whitelist was empty before, so there shouldn't be anyone with `StoredV15LandlineAccount` populated

apps/daimo-mobile/src/view/screen/send/SendTransferButton.tsx
andrewliu08: oops

apps/daimo-mobile/src/action/useLandlineDeposit.ts
andrewliu08: nvm, added offchain action for authentication

apps/daimo-mobile/src/storage/storedAccount.ts
```diff
@@ -63,6 +63,6 @@ export interface StoredV16Account extends StoredModel {
   exchangeRates: StoredV15CurrencyExchangeRate[];
   sentPaymentLinks: StoredV15DaimoLinkNoteV2[];
 
-  landlineSessionKey: string;
+  landlineSessionURL: string;
```
andrewliu08, line 66: @dcposch is it safe to do a field renaming on `storedAccount` like this?

PR #1295, opened 2024-08-27
db: add blocks to IndexDB setup

kayleegeorge:
- add `blocks` table
- readd `IndexDB` setup (got deleted in commit https://github.com/daimo-eth/daimo/commit/6ee66925468f77a9f6e8ab88a99691fa8ceb8911)

### DIFF
--- a/packages/daimo-api/src/db/db.ts
+++ b/packages/daimo-api/src/db/db.ts
@@ -184,12 +184,12 @@ export class DB {
         -- DaimoDB Index setup (filtered)
         --
 
-        CREATE TABLE IF NOT EXISTS index.blocks (
-          chain_id bigint NOT NULL,
-          number bigint NOT NULL,
-          hash bytea NOT NULL,
-          parent_hash bytea NOT NULL,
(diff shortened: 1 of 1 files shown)


packages/daimo-api/src/db/db.ts
```diff
@@ -181,8 +181,17 @@ export class DB {
...
 
         --
-        -- DaimoDB Index setup (filtered from Jumbo)
+        -- DaimoDB Index setup (filtered)
         --
+
+        CREATE TABLE IF NOT EXISTS index.blocks (
+          number bigint NOT NULL,
+          hash text NOT NULL,
+          parent_hash text NOT NULL,
```
dcposch, line 190: hash and parent_hash can be `bytea` to match how we're treating addresses

dcposch: + needs a `chain_id bigint NOT NULL`

dcposch (approved):
LGTM after comments

dcposch (approved):

PR #1299, opened 2024-08-30
common: add viem client common functions

kayleegeorge:
- common functions used for viem client setup


dcposch (approved):
LGTM

I think we can dedupe a bit with `chain.ts` etc, but can do that in future PR.

PR #1301, opened 2024-08-31
api: better logging and error handling for landline

andrewliu08:


packages/daimo-api/src/landline/connector.ts
```diff
@@ -35,22 +35,46 @@ export function getLandlineURL(daimoAddress: string, sessionKey: string) {
...
 ): Promise<LandlineSessionKey> {
-  // @ts-ignore
-  const sessionKey = await landlineTrpc.getOrCreateSessionKey.mutate({
-    daimoAddress,
-  });
-  return sessionKey;
+  console.log(`[LANDLINE] getting session key for ${daimoAddress}`);
+
+  try {
+    // @ts-ignore
```
dcposch, line 41: why does this require a ts-ignore?

dcposch (approved):
LGreatTM after comments

packages/daimo-api/src/landline/connector.ts
andrewliu08: the `landlineTrpc` is untyped rn since the trpc definition is in a separate repo. need to include ts-ignore otherwise the linter will say that the trpc method is unrecognized

PR #1304, opened 2024-09-03
api: better swap logging

dcposch:


packages/daimo-api/src/contract/foreignCoinIndexer.ts
```diff
@@ -80,8 +80,8 @@ export class ForeignCoinIndexer extends Indexer {
           ])
           .where("chain_id", "=", "" + chainConfig.chainL2.id)
           .where((eb) => eb.between("block_num", "" + from, "" + to))
-          .orderBy("block_num", "asc")
-          .orderBy("sort_idx", "asc")
```
dcposch, line 84: consistent sorting everywhere: (block_num, sort_idx | log_idx, depending on table)

...in each case, the second sort is ascending order across all txs within each block

andrewliu08 (commented):
lgtm

PR #1308, opened 2024-09-06
api, contract: support EntryPoint v0.6 & v0.7

dcposch:
- DAv1 uses v0.6
- DAv2 uses v0.7
- also, delete bundle compression. not needed now + doesn't support v0.7 yet.


packages/daimo-api/package.json
```diff
@@ -18,7 +18,6 @@
   "repository": "https://github.com/daimo-eth/daimo.git",
   "dependencies": {
     "@coinbase/cbpay-js": "^1.8.0",
-    "@daimo/bulk": "^0.3.1",
```
dcposch, line 21: RIP for now

packages/daimo-api/src/network/bundlerClient.ts
```diff
@@ -243,6 +175,23 @@ function packGasFees({
 function packHiLo(hi: bigint, lo: bigint) {
   return concatHex([
     numberToHex(hi, { size: 16 }),
-    numberToHex(lo, { size: 32 }),
+    numberToHex(lo, { size: 16 }),
```
dcposch, line 178: bundle compression no longer immediately necessary for two reasons:
- calldata is cheap for now, blob data priced @ nearly free
- ERC-4337 v0.7 has a less-egregious userop serialization format

dcposch:
shipping cc @andrewliu08 for post review

packages/daimo-contract/src/backcompat/entryPointV06.ts
```diff
@@ -0,0 +1,1013 @@
+export const entryPointV06Address =
+  "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789";
+export const entryPointV06ABI = [
```
andrewliu08, line 3: is there a corresponding codegen for `entryPointV07`?

packages/daimo-api/package.json
andrewliu08: what was daimo bulk? contracts for handling ops with the old entrypoint?

packages/daimo-api/src/network/bundlerClient.ts
andrewliu08, line 178: what is this change for?

andrewliu08 (approved):
lgtm if tested

packages/daimo-api/src/network/bundlerClient.ts
dcposch: fixing a bug. the correct serialization for the EntryPoint v0.7 PackedUserOp = in a few spots, needs to pack two uint128s into a single bytes32

packages/daimo-contract/src/backcompat/entryPointV06.ts
dcposch: yes, `entryPointABI` in the codegen output

I've pasted the old one into the `backcompat/` folder so we don't have to pull in both the new and old version of the `account-abstraction` forge dependency

PR #1342, opened 2024-09-23
mobile: fix cross-chain check

kayleegeorge:
Fixes the `isSwap` check for asserting the transfer is not a bridge+swap. 

Tokens have different addresses on different chains (e.g. Base USDC token addr != Arb USDC token addr), so must instead check that the `toCoin` address is not the same as the `bridgeCoin` for the `toChain` to check whether it's a swap


dcposch (approved):
LGTM

PR #1346, opened 2024-09-24
dev: upgrade deps

dcposch:
* Upgrade from wagmi v1 to v2
* Upgrade from rainbowkit v1 to v2


packages/daimo-api/package.json
```diff
@@ -18,9 +18,9 @@
   "repository": "https://github.com/daimo-eth/daimo.git",
   "dependencies": {
     "@coinbase/cbpay-js": "^1.8.0",
-    "@daimo/common": "*",
-    "@daimo/contract": "*",
-    "@daimo/userop": "*",
+    "@daimo/common": "0.2.10",
+    "@daimo/contract": "0.2.11",
+    "@daimo/userop": "0.1.0",
```
andrewliu08, line 23: just wondering, why do we need to specify the versions here if these are all in the same repo?

apps/daimo-mobile/package.json
```diff
@@ -85,9 +85,9 @@
     "immutable": "^5.0.0-beta.5",
     "jest": "^29.5.0",
     "jest-expo": "~50.0.3",
-    "react": "18.2.0",
+    "react": "*",
     "react-error-boundary": "^4.0.12",
-    "react-native": "0.73.4",
+    "react-native": "*",
```
andrewliu08, line 90: why set these to "*"?

dcposch: version defined in the root `package.json`

there was some version hell involving the exact version of react-native required by Expo

PR #1347, opened 2024-09-25
api: indexer debug health check

kayleegeorge:
- add block latencies to indexer debug health check


packages/daimo-api/src/db/indexWatcher.ts
```diff
@@ -179,6 +179,39 @@ export class IndexWatcher {
     return Number(result?.latest_block_num || 0);
   }
 
+  // Get the block write to postgres latency over the last numBlocks
+  async getLatency() {
+    // Get the block times and latencies for the last 1000 blocks
```
dcposch, line 184: 🔥

dcposch (approved):
LGreatTM!

PR #1361, opened 2024-10-21
daimo api: disable landline

andrewliu08:
Fixes issue https://github.com/daimo-eth/daimo-internal/issues/146

Remove landline from the deposit page but keep landline transfers in the transfer history

Simulator Screenshot - iPhone 15 Pro - 2024-10-21 at 16 35 48
Simulator Screenshot - iPhone 15 Pro - 2024-10-21 at 16 35 06


dcposch (approved):
LGreatTM

PR #1370, opened 2024-11-15
btc

nalinbhardwaj:
example:
Google Chrome 2024-11-15 08 39 33

tested end to end with two payments, didn't record video because of how long it takes lol


dcposch (commented):
LGTM

PR #1372, opened 2024-11-30
pay contract: new daimo pay relayer

andrewliu08:
- Changed daimo pay relayer to support multiple owners. This allows us to have separate EOAs to run starts and finishes while still using the same relayer contract.
- Modified the relayer contract to hold balances for tipping rather than having the relayer EOA hold this balance.
- Made `maxPreTip` and `maxPostTip` separate parameters. This could be useful to avoid unintentionally tipping when we switch to having the relayer hold multiple tokens
- `swapAndTip` now emits an event upon success. This will help us track how tipping affects our PnL

## Testing
Wrote test suite for new contract which tests:
1. only the contract creator can withdraw token balances from the relayer contract
2. `swapAndTip` pre-tip and post-tip work correctly
3. `swapAndTip` doesn't allow pre-tip or post-tip to exceed the max tip amount
4. excess swap output is kept by the contract


packages/contract/script/pay/DeployDaimoPayRelayer.s.sol
```diff
@@ -23,6 +23,10 @@ contract DeployDaimoPayRelayer is Script {
 
         console.log("daimoPayRelayer deployed at address:", daimoPayRelayer);
 
+        address relayer = 0x723A63fb50dA50A26997Fb99A2Eb151E4F8c5227;
```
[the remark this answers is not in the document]
andrewliu08: this is just the another EOA i created for starts/claims

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -2,83 +2,122 @@
...
-    constructor(address _owner) Ownable(_owner) {}
+    bytes32 public constant RELAYER_ROLE = keccak256("RELAYER_ROLE");
+
+    event SwapAndTip(
+        address indexed requiredTokenIn,
+        uint256 suppliedAmountIn,
+        address indexed requiredTokenOut,
+        uint256 swapAmountOut,
+        uint256 maxPreTip,
+        uint256 maxPostTip
```
dcposch, line 25: might be worth calling this `maxIntentTip` and `maxStartTip` for clarity

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -2,83 +2,122 @@
...
+    function withdrawAmount(
+        IERC20 token,
+        uint256 amount
+    ) public onlyRole(DEFAULT_ADMIN_ROLE) {
+        TokenUtils.transfer(token, payable(msg.sender), amount);
+    }
+
+    // Withdraws the full balance of a token from the contract to the admin.
+    function withdrawBalance(IERC20 token) public onlyRole(DEFAULT_ADMIN_ROLE) {
+        TokenUtils.transferBalance(token, payable(msg.sender));
```
dcposch, line 50: maybe have this return `uint256`, then `return TokenUtils.transferBalance...`

this way if admin needs to send the token balance on to anywhere else they know the amount cleanly

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -2,83 +2,122 @@
...
         TokenAmount calldata requiredTokenIn,
         uint256 suppliedTokenInAmount,
         TokenAmount calldata requiredTokenOut,
-        uint256 maxTip,
+        uint256 maxPreTip,
+        uint256 maxPostTip,
         Call calldata innerSwap
     ) external payable {
-        require(tx.origin == owner(), "DPR: only usable by owner");
+        require(hasRole(RELAYER_ROLE, tx.origin), "DPR: only relayer");
```
dcposch, line 68: 🚀

packages/contract/src/pay/DaimoPayRelayer.sol
andrewliu08: `maxStartTip` would be the post-tip and `maxIntentTip` would be the pre-tip?

personally, i prefer `maxPreTip` and `maxPostTip`. i think this naming makes more sense at the function-level context, whereas `maxStartTip` and `maxIntentTip` requires global-level context to make sense

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -2,83 +2,122 @@
...
+        //////////////////////////////////////////////////////////////
+        // SWAP
+        //////////////////////////////////////////////////////////////
+
+        // forceApprove() not necessary, we check correct tokenOut amount.
+        // Approve requiredTokenIn.amount even if it's greater than
+        // suppliedTokenInAmount. The difference is tipped by the contract. We
+        // already checked that the tip is within maxPreTip.
+        if (innerSwap.to != address(0)) {
+            requiredTokenIn.token.approve(innerSwap.to, requiredTokenIn.amount);
```
dcposch, line 117: forceapprove handles USDT inputs

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol#L78-L94

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -91,30 +130,34 @@ contract DaimoPayRelayer is Ownable2Step {
...
-        // Check the amount output from the swap. The contract owner tips the
-        // difference if needed. If there are excess tokens, transfer them to
-        // the owner.
+        //////////////////////////////////////////////////////////////
+        // POST-SWAP
+        //////////////////////////////////////////////////////////////
+
+        // If we received less than required, check that the amount we need to
+        // tip is within maxPostTip.
         if (swapAmountOut < requiredTokenOut.amount) {
```
dcposch, line 139: so if i understand correctly:

- during `start`, we have some amount of (random token) and need to produce an exact amount of the bridge token. maxPreTip is 0, maxPostTip is the max amount of the bridge token (eg USDC) we're willing to tip.

- during `fastfinish`/`claim` (whichever is actually running the intent) we have some amount of bridge token and we need an exact amount of destination token. maxPostTip is now 0, and maxPreTip is whatever amount of the bridge token we're willing to tip.

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -91,30 +130,34 @@ contract DaimoPayRelayer is Ownable2Step {
...
-        } else {
-            // Give excess tokens to owner.
-            uint256 tip = swapAmountOut - requiredTokenOut.amount;
-            TokenUtils.transfer(requiredTokenOut.token, payable(owner()), tip);
+            require(outShortfall <= maxPostTip, "DPR: excessive post tip");
         }
 
+        // Transfer the required output tokens to the caller, tipping the
+        // shortfall if needed. If there are surplus tokens from the swap, keep
+        // them.
```
dcposch, line 146: one effect of this is that the relayer will accumulate change in whatever the destination tokens are.

for ethOS, for example, it'll accumulate ETH on Base.

packages/contract/src/pay/DaimoPayRelayer.sol
```diff
@@ -124,7 +167,7 @@ contract DaimoPayRelayer is Ownable2Step {
         Call[] calldata startCalls,
         bytes calldata bridgeExtraData,
         Call[] calldata postCalls
-    ) public payable onlyOwner {
+    ) public payable onlyRole(RELAYER_ROLE) {
```
dcposch, line 170: it might be worth renaming this from `RELAYER_ROLE` / `relayer` etc to `RELAY_EOA_ROLE` / `relayEOA` or similar, to distinguish it from the relayer contract

packages/daimo-contract/src/chain.ts
```diff
@@ -72,6 +72,14 @@ export type DAv2Chain = {
   isTestnet?: boolean;
 };
 
+/**
+ * Returns whether a given chainID is an EVM chain.
+ * TODO: move Solana etc to their own ID space.
+ */
+export function isEvmChain(chainId: number) {
+  return chainId !== solana.chainId;
+}
```
dcposch, line 81: that's funny, I have this exact diff on another branch.

packages/contract/src/pay/DaimoPayRelayer.sol
andrewliu08: yes

dcposch (approved):
LGreatTM after comments

packages/contract/src/pay/DaimoPayRelayer.sol
andrewliu08: correct

PR #1387, opened 2025-02-07
@daimo/common 1.1.1: add disabledReason

dcposch:


andrewliu08:
we should update @daimo/contract to 1.1.1 to keep the versions in sync

